PR #4, opened 2015-02-13
Add Filter/Hook

dshanske:
Adds a filter to change the arguments passed to wp_insert_post and an action hook to perform actions after the post is updated using the micropub data.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -56,10 +56,10 @@ class Micropub {
     if (!array_key_exists('micropub', $wp->query_vars)) {
       return;
     }
-
+    $result = 0;
     $input = file_get_contents('php://input');
     parse_str($input, $q);
-
+    
     header('Content-Type: text/plain; charset=' . get_option('blog_charset'));
 
     Micropub::authorize($q);
@@ -91,7 +91,7 @@ class Micropub {
       }
 
       if ($q['action'] == 'edit' || !isset($q['action'])) {
-        Micropub::check_error(wp_update_post($args));
+        $result = Micropub::check_error(wp_update_post($args));
         status_header(204);
       } elseif ($q['action'] == 'delete') {
         Micropub::check_error(wp_trash_post($args['ID']));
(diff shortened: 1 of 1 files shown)


aaronpk:
Isn't there already a wordpress hook that runs after a post is updated? I would assume it would also run after this plugin updates the post.

dshanske:
But it wouldn't pass the micropub data to it.

snarfed:
right, that's the useful addition.

aaronpk:
cool! (I don't know very much about wordpress :smile:)

micropub.php
```diff
@@ -80,6 +79,7 @@ public static function parse_query($wp) {
     if (!isset($q['url']) || $q['action'] == 'create') {
       $args['post_status'] = 'publish';
       $result = Micropub::check_error(wp_insert_post($args));
+      $args['ID'] = $result;
```
snarfed: this can be removed

micropub.php
```diff
@@ -56,10 +56,9 @@ public static function parse_query($wp) {
     if (!array_key_exists('micropub', $wp->query_vars)) {
       return;
     }
-
     $input = file_get_contents('php://input');
     parse_str($input, $q);
-
+
```
snarfed: nit: whitespace

micropub.php
```diff
@@ -95,6 +95,7 @@ public static function parse_query($wp) {
         status_header(204);
       } elseif ($q['action'] == 'delete') {
         Micropub::check_error(wp_trash_post($args['ID']));
+	$result = $arg['ID']);
```
snarfed: this can be removed

micropub.php
```diff
@@ -111,8 +112,8 @@ public static function parse_query($wp) {
         exit;
       }
     }
-
-    do_action('micropub_request', $q);
+
```
snarfed: nit: whitespace

micropub.php
```diff
@@ -187,7 +188,7 @@ private static function map_params($q) {
       $args['post_date'] = iso8601_to_datetime($q['published']);
     }
 
-    return $args;
+    return apply_filter('before_micropub', $args, $q);
```
snarfed: mind moving this into `parse_query`, after the `map_params` call? it's subjective, but i'd prefer it there.

PR #8, opened 2015-02-14
Check for Theme Support of Microformats 2 and act accordingly

dshanske:
While the filter can override the work done by generate_post_content, the plugin marks up the content in a way that would be an issue if the theme was designed to mark it up already.

To address this, the theme should be able to say it is handling the markup by declaring,
add_theme_support('microformats2'); 

Will concurrently have my starter theme do this and propose same for SemPress.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -216,13 +216,12 @@ class Micropub {
     if (current_theme_supports('microformats2')) {
       if (isset($_POST['content']) {
         $args['post_content'] = $_POST['content']));
-	}
+      }
     }
     // Else markup the content before passing it through
     else {
       $args['post_content'] = Micropub::generate_post_content();
     }
-   
     return $args;
   }
 
--- a/micropub.php
+++ b/micropub.php
@@ -214,15 +214,14 @@ class Micropub {
     }
     // If the theme declares it supports microformats2, pass the content through
     if(current_theme_supports('microformats2')) {
-	if (isset($_POST['content']))
-	    {
+	if (isset($_POST['content']) {
 		$args['post_content'] = $_POST['content']));
-	    }
-        }
-    // Else markup the content before passing it through
-    else{
-    	$args['post_content'] = Micropub::generate_post_content();
 	}
+    }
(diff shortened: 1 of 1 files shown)


dshanske:
I know about my style issues. It drives pfefferle crazy too. I can't seem to match it.

I'm working on a new theme. But I'll put it into the current one now. It's just a single line.

As for the e-content class...since microformats2 themes wrap the_content in e-content, I would say that the way I set it up makes sense. If the theme supports it, disable marking up content entirely. The theme could add to the filter to address itself.

snarfed:
i see your point...but `generate_post_content` does a lot of valuable (micropub-specific) work that we should let themes reuse and not have to duplicate themselves. can we think of any alternatives?

dshanske:
Would have to think about that. I'm thinking this might be the best way for now.

snarfed:
hrm. themes are generally expected to just render pages, not usually write to the db, right? can you describe how you envision an mf2 compliant theme would handle a micropub request with eg in-reply-to, or h=event, and not lose data?

dshanske:
Well, ideally, micropub could store the extra pieces as metadata on the post so the theme could render it.

snarfed:
right, makes sense. is there a standard for that yet? maybe post kinds' or semantic linkbacks'? if we had a plan for that, I'd be more ok with this change.

dshanske:
Well, semantic linkbacks works on comments, not post data.

Post Kinds does have a metadata standard. It's an array stored in post meta. You don't need to support the taxonomy to use the same format.

response['url'] - The URL the post is responding to.
response['title'] - The title of same
response['author'] - The author
response['icon'] - The author icon/image/avatar
response['content'] - Content or a Citation...to allow for an excerpt for context.

The plugin doesn't require that any of them be in the array. It renders based on what is there.

The only context field I'm not currently supporting is publish date. But that would be response['date'] if I did.

snarfed:
done! bee68ff3c7de91173f9cf8ecddef580478fa9dec

PR #20, opened 2015-10-24
Comply with WordPress Coding Standards

dshanske:
I've had it sitting there for nearly 3 months. Might as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,109 @@
+# Travis CI (MIT License) configuration file.
+# @link https://travis-ci.org/
+
+# Declare project language.
+# @link http://about.travis-ci.org/docs/user/languages/php/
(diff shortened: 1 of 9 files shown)


snarfed:
cool, thanks for sending!

as discussed on irc, the main additional thing this needs is docs in the readme on what these all are and how contributors can use them.

snarfed:
quite a refactoring! looking forward to the tests too! :P

dshanske:
If you look at the way it is designed...the argument generation is off the filter or the hook. That makes it very easy to test or even replace the parts compared to the private functions you had before.

dshanske:
Going to close this and approach it in smaller pieces. I've also learned a lot since I wrote this with the webmentions refactoring.

PR #26, opened 2016-06-30
Move Pre and Post Processing Functions to Hooks

dshanske:
This changes the method of removing the generated post content, which I need removed for using Post Kinds, to a removing of a filter over a flag.

At the same time, it allows removing/replacing of upload handling, geodata saving, and mf2 storing to replace with custom functions as needed.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -61,6 +61,17 @@ class Micropub {
     add_action('send_headers', array('Micropub', 'http_header'));
     add_filter('host_meta', array('Micropub', 'jrd_links'));
     add_filter('webfinger_data', array('Micropub', 'jrd_links'));
+
+    // Add Markup to Post Content.
+    add_filter('before_micropub', array('Micropub', 'generate_post_content'));
+
+    // Store MF2 as post meta.
+    add_action('after_micropub', array('Micropub', 'store_mf2'));
+    // Sideload any provided photos.
+    add_action('after_micropub', array('Micropub', 'default_file_handler'));
+    // Store geodata in WordPress format.
+    add_action('after_micropub', array('Micropub', 'store_geodata'));
+
   }
 
   /**
@@ -129,7 +140,6 @@ class Micropub {
       kses_remove_filters();  // prevent sanitizing HTML tags in post_content
       $args['ID'] = Micropub::check_error(wp_insert_post($args));
       kses_init_filters();
-      Micropub::postprocess($args['ID']);
       status_header(201);
       header('Location: ' . get_permalink($args['ID']));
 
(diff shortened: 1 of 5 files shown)


snarfed:
cool!

have you gotten the tests running? do they pass with this?

dshanske:
I get a Strict Standard warning because you have two extra parameters in the wp_die_handler. And they do pass with this.

micropub.php
```diff
@@ -310,18 +319,12 @@ private static function generate_args() {
         }
       }
     }
-    // If the theme declares it supports microformats2, pass the content through
-    if (current_theme_supports('microformats2')) {
```
snarfed: was this removed intentionally? seems like we still want it, right?

micropub.php
```diff
@@ -413,17 +417,23 @@ private static function generate_event() {
...
       require_once(ABSPATH . 'wp-admin/includes/media.php');
       Micropub::check_error(media_handle_upload('photo', $post_id));
     }
+  }
 
+  /**
+   * Adds geodata to post.
+   *
+   */
+  private static function geodata($post_id) {
```
snarfed: minor nit: add a verb to this function name? :P

snarfed:
this looks great otherwise, thank you for doing it!

funny also that the parts it touches are the parts we don't have tests for yet. :P maybe consider adding those tests in your next PR!

dshanske:
The add theme support was removed intentionally. The filter removal is cleaner. Because in my case, my plugin is removing it, not a theme. More universal.

dshanske:
I should be able to mirror your tests and try to test for mf2 priorities.

dshanske:
I will push a second commit for the name issue. And will see about testing one or more tonight.

snarfed:
> The add theme support was removed intentionally. The filter removal is cleaner. Because in my case, my plugin is removing it, not a theme. More universal.

hmm. so you mean, plugins and themes that implement/support mf2 should remove `generate_post_content` from the `before_micropub` filter? that means they all need to know the micropub plugin exists, and that they need to do that, which seems error prone. i'm happy with moving it to a filter, but how about we keep the `current_theme_supports` check too?

snarfed:
thanks! you'll need to pull from upstream and rebase this branch and resolve the conflicts, then we can merge!

dshanske:
Didn't notice you'd committed. Sorry.

snarfed:
np, thanks! and nice to see circle running the tests and that they pass too.

mind squashing the commits now?

dshanske:
Will have to wait till tomorrow.

dshanske:
I hope I got that right.

snarfed:
thanks! the squash part looks OK, but this branch isn't rebased onto master cleanly. you can tell when you look at the bottom of the PR, below this comment, there's an error message that says there are conflicts that need to be resolved.

looks like the problem is that you introduced a merge commit, probably when you pulled master into your branch at some point. you don't usually want to do that; you generally want to rebase your branch onto master instead. sounds like you'd looked at some tutorials on this before, maybe check them again?

in the meantime, the easiest way to recover here may be to close this PR, start a new branch, and cherry pick just your one squashed commit.

PR #27, opened 2016-07-04
Move Processing to Hooks and Filters REDUX

dshanske:
Allows removing/replacing of upload handling, geodata saving, and mf2 storing to replace with custom functions as needed.

The difference between this and #26 is that this switches it from post-processing to pre-processing. The advantage to this is that as opposed to just removing the data...still an option...it can be transformed by a filter later in the process.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -62,7 +62,7 @@ class Micropub {
     add_filter('host_meta', array('Micropub', 'jrd_links'));
     add_filter('webfinger_data', array('Micropub', 'jrd_links'));
 
-    // Store MF2
+    // Store MF2 - set at priority 8 in order to occur before the other filters.
     add_filter('before_micropub', array('Micropub', 'store_mf2'), 8);
     // Postprocess
     add_filter('before_micropub', array('Micropub', 'generate_post_content'));
@@ -337,7 +337,8 @@ class Micropub {
     if (current_theme_supports('microformats2')) {
       return $args;
 		}
-    // Disable if Kind Taxonomy is enabled
+    // Disable if Kind Taxonomy is enabled. The Kind Taxonomy is used in the Post Kinds plugin.
+    // The Post Kinds plugin handles its own markup. 
     if (taxonomy_exists('kind')) {
       return $args;
     }
--- a/micropub.php
+++ b/micropub.php
@@ -5,7 +5,7 @@
  Description: <a href="https://indiewebcamp.com/micropub">Micropub</a> server.
  Author: Ryan Barrett
  Author URI: https://snarfed.org/
- Version: 0.4
+ Version: 0.5
 */
 
 // Example command line for testing:
@@ -61,7 +61,16 @@ class Micropub {
     add_action('send_headers', array('Micropub', 'http_header'));
     add_filter('host_meta', array('Micropub', 'jrd_links'));
     add_filter('webfinger_data', array('Micropub', 'jrd_links'));
-  }
+
+    // Store MF2
+    add_filter('before_micropub', array('Micropub', 'store_mf2'), 8);
+    // Postprocess
+    add_filter('before_micropub', array('Micropub', 'generate_post_content'));
+    // Store Geodata
+    add_filter('before_micropub', array('Micropub', 'store_geodata'));
+    // Sideload any provided photos.
+    add_action('after_micropub', array('Micropub', 'default_file_handler'));
+	}
 
   /**
    * Adds some query vars
@@ -115,7 +124,6 @@ class Micropub {
       $_POST['action'] = isset($_POST['operation']) ? $_POST['operation']
         : isset($_POST['url']) ? 'edit' : 'create';
     }
-
     $args = apply_filters('before_micropub', Micropub::generate_args());
     if ($user_id) {
       $args['post_author'] = $user_id;
@@ -129,7 +137,6 @@ class Micropub {
       kses_remove_filters();  // prevent sanitizing HTML tags in post_content
       $args['ID'] = Micropub::check_error(wp_insert_post($args));
       kses_init_filters();
-      Micropub::postprocess($args['ID']);
       status_header(201);
       header('Location: ' . get_permalink($args['ID']));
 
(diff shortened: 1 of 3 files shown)


micropub.php
```diff
@@ -61,7 +61,16 @@ public static function init() {
     add_action('send_headers', array('Micropub', 'http_header'));
     add_filter('host_meta', array('Micropub', 'jrd_links'));
     add_filter('webfinger_data', array('Micropub', 'jrd_links'));
-  }
+
+    // Store MF2
+    add_filter('before_micropub', array('Micropub', 'store_mf2'), 8);
```
snarfed: is the priority 8 here important?

dshanske: Default priority is 10. Storing the MF2 in the meta should occur before the others, so I set it to 8. That means other filters will have access to it in the args array.

micropub.php
```diff
@@ -310,18 +316,14 @@ private static function generate_args() {
...
     }
-    // If the theme declares it supports microformats2, pass the content through
-    if (current_theme_supports('microformats2')) {
-      if (isset($_POST['content'])) {
-        $args['post_content'] = $_POST['content'];
-      }
-      else if (isset($_POST['summary'])) {
-        $args['post_content'] = $_POST['summary'];
-      }
+    if (!isset($args['meta_input'])) {
```
snarfed: `meta_input` isn't used in this method, so i think you can remove this block.

micropub.php
```diff
@@ -432,18 +452,21 @@ private static function postprocess($post_id) {
       $geo = explode(':', $geo);
       $geo = explode(';', $geo[0]);
       $coords = explode(',', $geo[0]);
-      update_post_meta($post_id, 'geo_latitude', trim($coords[0]));
-      update_post_meta($post_id, 'geo_longitude', trim($coords[1]));
+      $args['meta_input']['geo_latitude'] = trim($coords[0]);
```
snarfed: if we want to delay storing the post meta, that's ok, but we do still need to store it at some point. it looks like you removed the `updated_post_meta` alls here but didn't add them anywhere else.

snarfed: you might consider adding a few unit tests in this PR, e.g. checking that post meta is stored. :P

micropub.php
```diff
@@ -452,14 +475,14 @@ private static function store_mf2($post_id) {
       }
       if (!in_array($key, $blacklist)) {
         $key = 'mf2_' . $key;
-        delete_post_meta($post_id, $key);  // clear old value(s)
         foreach ($value as $val) {
           if (!empty($val)) {
-            add_post_meta($post_id, $key, $val);
+            $args['meta_input'][$key] = $val;
```
snarfed: ditto the removed `delete_post_meta` and `add_post_meta` calls here.

micropub.php
```diff
@@ -330,7 +332,15 @@ private static function generate_args() {
    * Generates and returns a post_content string suitable for wp_insert_post()
    * and friends.
    */
-  private static function generate_post_content() {
+  public static function generate_post_content($args) {
+    // If the theme declares it supports microformats2, pass the content through
+    if (current_theme_supports('microformats2')) {
+      return $args;
+		}
+    // Disable if Kind Taxonomy is enabled
```
snarfed: i wonder if there's a way to do this without special casing the post kinds plugin explicitly like this?

micropub.php
snarfed: sgtm. probably worth adding a comment.

micropub.php
dshanske: I wanted it created by default, but point taken.

micropub.php
dshanske: I could move that code back to the Post Kinds plugin. I had it there originally.

micropub.php
dshanske: wp_insert_post, which is used in new and updated post entries, takes anything stored in meta_input and updates the  post meta already. That is why I switched it. It has advantages and removes some code.

Also, checked...it is being stored.

micropub.php
dshanske: deleting and then adding isn't necessary. See above.

snarfed:
if you're bumping the version now, mind starting a version 0.5 section in readme.txt?

micropub.php
snarfed: > The feature wasn't added till 4.4 though, which bumps the minimum version number to 4.4. But no one should be running a version older than 2 releases back anyway and be supported.

tough talk! :P

right now we only require >=3.0.1: https://wordpress.org/plugins/micropub/ . bumping that up to 4.4 is a big jump, but i'm ok with it if you think we should.

snarfed: > Also, checked...it is being stored.

manual checks are fleeting. unit tests are forever. :P

micropub.php
snarfed: got it. i don't feel strongly, and you know the wordpress conventions better than me. whatever you think is best!

dshanske:
I agree it is a tough stance. I will try unit tests, still bad at it. Will add a readme. Also, as for conventions, two versions back is what most projects guarantee for Core.

snarfed:
> two versions back is what most projects guarantee for Core.

aha! so then 4.4 is maybe a bit too aggressive, since it's only one version back.

dshanske:
4.6 will be out shortly.

dshanske:
I added a unit test.

micropub.php
```diff
@@ -337,7 +337,8 @@ public static function generate_post_content($args) {
     if (current_theme_supports('microformats2')) {
       return $args;
 		}
-    // Disable if Kind Taxonomy is enabled
+    // Disable if Kind Taxonomy is enabled. The Kind Taxonomy is used in the Post Kinds plugin.
+    // The Post Kinds plugin handles its own markup.
```
snarfed: the valuable part to add here is the link, eg https://wordpress.org/plugins/indieweb-post-kinds/

micropub.php
```diff
@@ -62,7 +62,7 @@ public static function init() {
     add_filter('host_meta', array('Micropub', 'jrd_links'));
     add_filter('webfinger_data', array('Micropub', 'jrd_links'));
 
-    // Store MF2
+    // Store MF2 - set at priority 8 in order to occur before the other filters.
```
snarfed: the valuable thing to comment here is *why* we want it to occur before the other filters. :P

snarfed:
awesome! nice work on the test! merging now. i'm going to convert this to wordpress style, so i'll also tweak the comments in that commit.

congrats!

PR #30, opened 2016-07-07
Rearrange Functions to Call Query Before Handling

dshanske:
This is an attempt to start separating functions before working on updating the endpoint for queries, including the syndication one...the primary one I wish to address.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -92,9 +92,6 @@ class Micropub {
 	 * Parse the micropub request and render the document
 	 *
 	 * @param WP $wp WordPress request context
-	 *
-	 * @uses apply_filter() Calls 'before_micropub' on the default request
-	 * @uses do_action() Calls 'after_micropub' for additional postprocessing
 	 */
 	public static function parse_query( $wp ) {
 		if ( ! array_key_exists( 'micropub', $wp->query_vars ) ) {
@@ -103,7 +100,6 @@ class Micropub {
 		if ( WP_DEBUG ) {
 			error_log( 'Micropub Data: ' . serialize( $_GET ) . ' ' . serialize( $_POST ) );
 		}
-		header( 'Content-Type: text/plain; charset=' . get_option( 'blog_charset' ) );
 		// For debug purposes be able to bypass Micropub auth with WordPress auth
 		if ( MICROPUB_LOCAL_AUTH ) {
 			if ( ! is_user_logged_in() ) {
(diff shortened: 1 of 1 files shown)


micropub.php
```diff
@@ -114,17 +110,94 @@ public static function parse_query( $wp ) {
 		else {
 			$user_id = Micropub::authorize();
 		}
+		// If q is set this must be a query so 
+		if ( isset( $_GET['q'] ) ) {
+			Micropub::query_handler( $user_id );  // calls wp_die
+			wp_die();
```
snarfed: you can remove this, since `query_handler` calls it. and probably put `form_handler` in an `else`.

micropub.php
```diff
@@ -114,17 +110,94 @@ public static function parse_query( $wp ) {
...
+		}
+		// This handles the form encoded requests.
+		// TODO: Rewrite Form Handler and add JSON Handler
+		self::form_handler( $user_id );
+		// The handlers should exit before ever hitting the wp_die
+		wp_die();
```
snarfed: probably just remove this too

micropub.php
```diff
@@ -114,17 +110,94 @@ public static function parse_query( $wp ) {
 		else {
 			$user_id = Micropub::authorize();
 		}
+		// If q is set this must be a query so
```
snarfed: minor nit: "so..."? alternatively, this code is self explanatory, so consider just removing this comment and the one on line 118 altogether. commenting is like goldilocks, you want just enough! :P (also, nit, trailing whitespace.)

micropub.php
```diff
@@ -114,17 +110,94 @@ public static function parse_query( $wp ) {
...
+		wp_die();
+	}
+
+	/**
+	 * Use tokens.indieauth.com to validate the access token.
+	 *
+	 * If the token is valid, returns the user id to use as the post's author, or
+	 * NULL if the token only matched the site URL and no specific user.
+	 */
+	private static function authorize() {
```
snarfed: just fyi, this PR's diff is misleading and looks a lot bigger than it is because you moved this function in the file. if you leave it where it is instead, after `form_handler`, the diff is way smaller and more readable, and better represents the actual changes, which makes the PR easier to understand and review, and also more readable later.

not a big deal, no need to change it now. just a heads up for the future.

snarfed:
lgtm otherwise! happy to merge after those minor nits are fixed.

dshanske:
That better?

snarfed:
thanks, looks great!

Issue #35, opened 2016-09-07
"Invalid token" sends HTTP status code "200 OK"


cweiske:
This only happens when I'm using it offline without internet access:

```
$ curl -i -X POST -H 'Authorization: Bearer abc' --data 'h=entry&like-of=http%3A%2F%2Fanoweco.bogo%2F' 'http://wordpress.bogo/?micropub=endpoint'HTTP/1.1 200 OK
Date: Wed, 07 Sep 2016 06:40:40 GMT
Server: Apache/2.4.7 (Ubuntu)
X-Powered-By: PHP/5.5.9-1ubuntu4.19
Link: <http://wordpress.bogo/?micropub=endpoint>; rel="micropub"
Link: <http://anoweco.bogo/auth.php>; rel="authorization_endpoint"
Link: <http://anoweco.bogo/token.php>; rel="token_endpoint"
Content-Length: 24
Content-Type: text/plain; charset=UTF-8

invalid access token: 
```

(there is no message. when I'm online, there is a message and the correct status code is there)

snarfed:
thanks for the reports! i don't use this plugin myself, sadly, so i'm often the wrong person to work on it, but i may still take a look. @dshanske also often works on it.

PR #37, opened 2016-09-07
Use configured token endpoint instead of tokens.indieauth.com

cweiske:
This solves #35.


snarfed:
lgtm! thanks for the contribution.

btw, we're missing unit tests for authorization right now, but i'd love to have them eventually. feel free to try adding some if you're interested!

PR #38, opened 2016-09-07
Compare authorization header case-insensitive

### LINKED ISSUE #34: HTTP "authorization" header not accepted (case-sensitive)
HTTP headers are meant to be processed case-insensitive, but the code only accepts `Authorization` (word case) and not `authorization` (lowercase).

cweiske:
Resolves: #34

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -128,8 +128,11 @@ class Micropub {
 	private static function authorize() {
 		// find the access token
 		$headers = getallheaders();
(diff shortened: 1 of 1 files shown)


snarfed:
lgtm since this is the only place we look at the request headers. thanks!

voxpelli:
Seems like the WordPress way of doing it would be to just do `$_SERVER['HTTP_AUTHORIZATION']` – at least when looking in WordPress Core

PR #44, opened 2016-09-16
some fixes

cweiske:


snarfed:
thanks for the patch. ahh, late static binding, much better than `get_called_class`. is that a recent PHP feature though? does it reduce our backward compatibility? ok if so, just curious.

thanks for the content[html] fix too. mind adding a unit test? hopefully shouldn't be long.

cweiske:
Late Static Binding had been introduced in PHP 5.3 - http://php.net/lsb

the tests .. didn't try them at all yet.

cweiske:
- test for content[html] added
- "normal" non-html content gets htmlspecialchar'd now.

snarfed:
thanks again for the new test! lgtm.

looks like the test failure was because circle was using ubuntu 12, which has old versions of php and phpunit compared to what we both have locally. i switched us to ubuntu 14, https://circleci.com/docs/build-image-trusty/ , and that fixed it.

Issue #32, opened 2016-07-07
fix style issues reported by PHPCS


snarfed:
_very_ low priority, since the big style stuff is ok now - tabs for indents, spaces inside parens, braces around `if`/`elseif` blocks.

[PHPCS details in the readme.](https://github.com/snarfed/wordpress-micropub#development) [WordPress coding standards here.](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards)

snarfed:
hey @dshanske, to start, would you mind pasting in the full current PHPCS output here? i haven't set it all up, but i'd love to see the results!

dshanske:
[phpcsmicropub.txt](https://github.com/snarfed/wordpress-micropub/files/353575/phpcsmicropub.txt)

dshanske:
Figured that before I started messing around, might as well address this as it causes major changes in file structure.

snarfed:
we looked at incorporating the PHPCS style suggestions in #51, and it didn't go so well. way too many false positives, not many actually useful suggestions. i'm going to tentatively close this.

PR #51, opened 2017-01-07
WordPress Coding Standards

dshanske:
Attempts to address #32.

PHPCS identified issues with the endpoint URLs and how they were generated. Cleaned up header generation code to match similar changes in webmentions plugin and abstracted them into their own functions with filters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,14 @@ sideloaded from URLs.
 
 #### Filters and hooks
 
-Adds one filter: `before_micropub( $input )`
+Adds several filters: `before_micropub( $input )`
 
 Called before handling a Micropub request. Returns $input, possibly modified.
+
+`micropub_endpoint( $default )`, `micropub_authentication_endpoint( $default )`, `micropub_token_endpoint( $default)`
+
+Allows these three default endpoint locations to be filtered.
+
 
 ...and one hook: `after_micropub( $input, $wp_args = null)`
 
@@ -45,6 +50,8 @@ Arguments:
   arguments passed to wp_insert_post or wp_update_post. For deletes and
   undeletes, args['ID'] contains the post id to be (un)deleted. Null for queries.
 
+`$default`: The default endpoint location used by the plugin. The Authentication and Token Endpoints default to [IndieAuth](https://indieauth.com/), the Micropub endpoint to the `/?micropub=endpoint`
+
 #### Other
 
 Stores [microformats2](http://microformats.org/wiki/microformats2) properties in
@@ -60,7 +67,7 @@ won't matter, but just for the record.)
 ### Authentication and authorization
 
 Supports the full OAuth2/IndieAuth authentication and authorization flow.
-Defaults to IndieAuth. Custom auth and token endpoints can be used by overriding
+Defaults to [IndieAuth](https://indieauth.com/). Custom auth and token endpoints can be used by overriding
 the `MICROPUB_AUTHENTICATION_ENDPOINT` and `MICROPUB_TOKEN_ENDPOINT` endpoints.
 If the token's `me` value matches a WordPress user's URL, that user will be
 used. Otherwise, the token must match the site's URL, and no user will be used.
@@ -124,6 +131,7 @@ To set up PHPCodesniffer to test changes for adherance to WordPress Coding Stand
 #### 1.1 (unreleased)
 * Support [`h-adr`](http://microformats.org/wiki/h-adr), [`h-geo`](http://microformats.org/wiki/h-geo), and plain text values for `p-location`.
 * Bug fix for create/update with content[html].
+* Add filters for the three endpoints to allow override by plugin.
 
 #### 1.0.1
 * Remove accidental dependence on PHP 5.3 (#46).
--- a/micropub.php
+++ b/micropub.php
@@ -817,7 +817,7 @@ class Micropub {
 	}
 
 	public static function get_token_endpoint() {
-		return apply_filters( 'micropub_endpoint', MICROPUB_TOKEN_ENDPOINT );
+		return apply_filters( 'micropub_token_endpoint', MICROPUB_TOKEN_ENDPOINT );
 	}
 
 	/**
--- a/micropub.php
+++ b/micropub.php
@@ -61,8 +61,6 @@ if ( ! defined( 'MICROPUB_DRAFT_MODE' ) ) {
 	define( 'MICROPUB_DRAFT_MODE', '0' );
 }
 
-if ( ! class_exists( 'Micropub' ) ) :
-
 add_action( 'init', array( 'Micropub', 'init' ) );
 
 /**
@@ -72,8 +70,8 @@ class Micropub {
 	// associative array
 	public static $request_headers;
 
-    // associative array, read from JSON or form-encoded input. populated by
-    // load_input().
+	// associative array, read from JSON or form-encoded input. populated by
+	// load_input().
 	protected static $input;
 
 	/**
@@ -96,7 +94,7 @@ class Micropub {
 		add_filter( 'webfinger_data', array( $cls, 'jrd_links' ) );
 	}
 
-	/**
+		/**
 	 * Adds some query vars
 	 *
 	 * @param array $vars
@@ -107,7 +105,7 @@ class Micropub {
 		return $vars;
 	}
 
-	/**
+		/**
 	 * Parse the micropub request and render the document
 	 *
 	 * @param WP $wp WordPress request context
@@ -122,7 +120,7 @@ class Micropub {
 		static::load_input();
 		if ( WP_DEBUG ) {
 			error_log( 'Micropub Data: ' . serialize( $_GET ) . ' ' .
-					   serialize( static::$input ) );
+			serialize( static::$input ) );
 		}
 		static::$input = apply_filters( 'before_micropub', static::$input );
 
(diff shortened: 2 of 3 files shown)


snarfed:
thanks for the contribution!

i'm fine with the endpoint URL changes, but we should probably drop the indentation changes. i know PHPCS complained, but it's imperfect, like any automated tool, and in this case i think better judgment says no.

i try to break lines at around 80 chars when possible, since my (and many others') editor windows aren't much wider than that. PHP's coding standards agree: http://pear.php.net/manual/en/standards.indenting.php#standards.indenting

adding a 4 char tab to every line in the file reduces the horizontal real estate we have to work with, and arguably adds no value, since the `class_exists` isn't meaningful or important to remember when writing code inside this file.

also, neither wordpress's nor PHP's coding standards say anything about this explicitly:
* https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation
* http://pear.php.net/manual/en/standards.php


there's precedent for not indenting after a check in wordpress itself:
* https://github.com/WordPress/WordPress/blob/master/wp-includes/class-json.php#L2
* https://github.com/WordPress/WordPress/blob/master/wp-includes/class-simplepie.php#L2

and there's also precedent for skipping the `class_exists` check altogether in other wordpress plugins (including yours! :P):
* https://github.com/acegiak/Semantic-Linkbacks/blob/master/handler/semantic-linkbacks-microformats-handler.php
* https://github.com/dshanske/syndication-links/blob/master/includes/class-hcard-widget.php

i'd be fine with either removing the check or dropping the commit that adds the indentation from this PR. up to you!

micropub.php
```diff
@@ -810,31 +810,44 @@ private static function check_error( $result ) {
...
+		}
+
 		/**
 	 * The micropub autodicovery meta tags
 	 */
 		public static function html_header() {
-			echo '<link rel="micropub" href="' . site_url( '?micropub=endpoint' ) . '">';
-			echo '<link rel="authorization_endpoint" href="' . MICROPUB_AUTHENTICATION_ENDPOINT . '">';
-			echo '<link rel="token_endpoint" href="' . MICROPUB_TOKEN_ENDPOINT . '">';
+			printf( '<link rel="micropub" href="%s" />' . PHP_EOL, static::get_micropub_endpoint() );
```
snarfed: minor nit: `link` has no content, so you can drop the closing `/`s.

micropub.php
```diff
@@ -810,31 +810,44 @@ private static function check_error( $result ) {
 			return $result;
 		}
 
+
+		public static function get_micropub_endpoint() {
+			return apply_filters( 'micropub_endpoint', site_url( '?micropub=endpoint' ) );
```
snarfed: three new filters! cool! can you please add them to `README.md` and `readme.txt`, in both the _Filters and hooks_ sections and the changelogs?

snarfed (changes requested):

dshanske:
I'm going with removing the check since it would throw an error on activation or deactivation of a competing name only. And since this is the first Micropub plugin for WordPress....

micropub.php
dshanske: I don't have a strong opinion on either side of the / debate(it seems there are differing opinions on it. I liked the changes that @pfefferle made to similar entries in the webmention plugins because they are cleaner and addressed another issue, and he had the / .

micropub.php
dshanske: I will do that. Having the three different versions come from a function as opposed to hard-coded ensured that they'd always be set identically. And as long as there was a function required, the filter made sense.

micropub.php
snarfed: oh huh, i didn't realize the closing `/` was allowed (optional) in empty elements: https://www.w3.org/TR/html/syntax.html#start-tags

i'm fine either way then.

micropub.php
```diff
@@ -810,31 +810,44 @@ private static function check_error( $result ) {
...
+		}
+
+		public static function get_authentication_endpoint() {
+			return apply_filters( 'micropub_authentication_endpoint', MICROPUB_AUTHENTICATION_ENDPOINT );
+		}
+
+		public static function get_token_endpoint() {
+			return apply_filters( 'micropub_endpoint', MICROPUB_TOKEN_ENDPOINT );
```
snarfed: oh also, looks like this filter name needs to be `micropub_token_endpoint`

snarfed: ...which is a good sign that it's probably time for unit tests for these headers. :P consider adding some in this PR?

dshanske: Making the fixes now.

readme.txt
```diff
@@ -90,9 +96,9 @@ These configuration options can be enabled by adding them to your wp-config.php
 * `define('MICROPUB_LOCAL_AUTH', '1')` - Bypasses Micropub authentication in
 favor of WordPress authentication
 * `define('MICROPUB_AUTHENTICATION_ENDPOINT', 'https://indieauth.com/auth')` -
-Define a custom authentication endpoint
+Define a custom authentication endpoint (can also be overridden by filter)
```
snarfed: minor nit: consider mentioning the name of the filter :P (also below)

README.md
```diff
@@ -25,10 +25,15 @@ sideloaded from URLs.
 
 #### Filters and hooks
 
-Adds one filter: `before_micropub( $input )`
+Adds several filter: `before_micropub( $input )`
```
snarfed: minor nit, typo: filter*s*

snarfed:
thanks for the updates! some of the other style changes seem surprising too, but i'll check the wordpress and php standards and bite my tongue if they're all explicitly recommended. :P

otherwise, maybe consider a couple things:
* squash the two style commits? the indentation changes and then rollbackas are both pretty invasive, which causes problems when git annotating and following code changes across commits. this might help: https://help.github.com/articles/using-git-rebase-on-the-command-line/
* unit tests for the headers and filters :P

README.md
```diff
@@ -45,6 +50,8 @@ Arguments:
   arguments passed to wp_insert_post or wp_update_post. For deletes and
   undeletes, args['ID'] contains the post id to be (un)deleted. Null for queries.
 
+`$default`: The default endpoint location used by the plugin. The Authentication and Token Endpoints default to IndieAuth, the Micropub endpoint to the /?micropub=endpoint
```
snarfed: minor nits:
* please link IndieAuth (ie `[IndieAuth](https://indieauth.com/)`)
* _the /?micropub=endpoint_ => _\`/?micropub=endpoint\`_.

also in readme.txt.

dshanske:
I'll squash them. I was in the process of figuring out how to test a header.

micropub.php
```diff
@@ -122,7 +120,7 @@ public static function parse_query( $wp ) {
 		static::load_input();
 		if ( WP_DEBUG ) {
 			error_log( 'Micropub Data: ' . serialize( $_GET ) . ' ' .
-					   serialize( static::$input ) );
+			serialize( static::$input ) );
```
snarfed: please also revert these changes that remove indentation from line continuations. the indentation is needed for readability, and both wordpress's and PHP's code standards agree:

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#indentation
http://pear.php.net/manual/en/standards.funcalls.php#standards.funcalls.splitfunccall

again, not sure what PHPCS is smoking. :P

micropub.php
```diff
@@ -155,53 +153,55 @@ private static function authorize() {
...
+			$code, 'invalid access token: ' . $body );
 		} elseif ( ! in_array( 'post',  $scopes ) && ! in_array( 'create',  $scopes ) ) {
 			static::handle_authorize_error(
-				401, 'access token is missing post or create scope; got `' . $params['scope'] . '`' );
+			401, 'access token is missing post or create scope; got `' . $params['scope'] . '`' );
 		}
 
-		parse_str( $body, $resp );
-		$me = untrailingslashit( $resp['me'] );
+			parse_str( $body, $resp );
```
snarfed: these are wrong. also below. please revert.

micropub.php
```diff
@@ -212,23 +212,23 @@ public static function post_handler( $user_id ) {
...
 			if ( $user_id && ! user_can( $user_id, 'publish_posts' ) ) {
 				static::error( 403, 'user id ' . $user_id . ' cannot publish posts' );
 			}
 			$args = static::create( $user_id );
 			$status = 201;
 			static::header( 'Location', get_permalink( $args['ID'] ) );
 
-		// update
-		} elseif ( $action == 'update' || ! $action ) {
+			// update
```
snarfed: same, revert, and below

micropub.php
```diff
@@ -805,56 +805,65 @@ private static function check_error( $result ) {
...
-				} elseif ( $key == 'h' ) {
-					static::$input['type'] = array( 'h-' . $val );
-				} elseif ( $key == 'access_token' ) {
-					continue;
-				} else {
-					if ( ! isset( static::$input['properties'] ) ) {
-						static::$input['properties'] = array();
-					}
-					static::$input['properties'][ $key ] =
+		public static function jrd_links( $array ) {
```
snarfed: please squash these out entirely too. thanks!

micropub.php
```diff
@@ -696,14 +699,14 @@ public static function store_geodata( $args ) {
 				$coords = explode( ',', $geo[0] );
 				$args['meta_input']['geo_latitude'] = trim( $coords[0] );
 				$args['meta_input']['geo_longitude'] = trim( $coords[1] );
-			} elseif ( substr( $location, 0, 4 ) != 'http' ) {
+			} elseif ( substr( $location, 0, 4 ) !== 'http' ) {
```
snarfed: also switch order to make this yoda style like the others?

snarfed (changes requested):
thanks for squashing!

i looked at the rest of the style changes, and honestly, PHPCS seems pretty flawed at best. switching to "yoda style" conditionals and the few missing spaces inside parens are good, but switching to `===` and `!==` doesn't really matter in any of our cases, and all of the indentation changes were wrong. ah well!

snarfed:
honestly, it might be easier to start over and just do the endpoint and yoda style conditional changes, instead of trying to extract them out of these commits. up to you!

dshanske:
I thought it was easier to address the coding standard thing because it would move everything around. I think it doesn't necessarily help much. I may just leave that part for now, close this, and focus on adding functionality in a separate PR.

Issue #67, opened 2017-07-15
Storing token information with each post


lostfocus:
So, I've been thinking about finding a way of storing the token with each token - or at least the client_id. Am I missing something or would that require a pretty substantial rewrite?

snarfed:
interesting! definitely possible. I'm curious why you'd want to though... what's your use case?

lostfocus:
Mostly because I saw that @aaronpk is using a little "posted using [client]" line on his site and thought that'd be pretty nifty to have, too.

snarfed:
ah. sure!

i see client_id in the auth flow, https://indieauth.com/developers, but that's between the micropub client and the user's sure/auth endpoint. i don't see client_id in micropub itself, https://www.w3.org/TR/micropub/, and this plugin isn't an auth endpoint, so it can't decode tokens. (and they're implementation-dependent anyway.)

@aaronpk's site is its own auth endpoint, so he has access to client_id. I'm not sure what we can do here.

aaronpk:
How does this plugin verify access tokens?

lostfocus:
Hm, at least with indieauth.com as a token endpoint the client_id should be in the token response. (Like 'me' which is used in [line 184](https://github.com/snarfed/wordpress-micropub/blob/v1.2/micropub.php#L184).) It might be possible to do something there.

snarfed:
ooh, ok! we could definitely optionally grab client_id from the verify response if it's there.

(@aaronpk it does the standard indieauth verification, but the endpoint is user-configurable.)

@lostfocus interested in tackling this? should be straightforward!

PR #68, opened 2017-07-16
Store the indieauth token endpoint return as post metadata

lostfocus:
This should do it. ("it" being #67)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,10 @@ To set up PHPCodesniffer to test adherence to [WordPress Coding Standards](https
 
 ### Changelog
 
+#### 1.3 (unreleased)
+* Saves [access token response](https://tokens.indieauth.com/) in a post meta field `micropub_auth_response`.
+
+
 #### 1.2 (2017-06-25)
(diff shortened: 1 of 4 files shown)


snarfed:
awesome, thanks for doing this!

mind adding a test and updating the changelogs in the readmes? (you'd add a new _1.3 (unreleased)_ section.)

also, `token_data` is a pretty generic name for the post_meta key. let's call it `micropub_auth_response` instead. alternatively you could just pull out `client_id` and store it in `micropub_client_id`. i'm ok with either.

thanks again!

lostfocus:
Yeah, for the tests I'll have to re-figure out the whole phpunit/WP test/… things first, so that might take a while. It has been a while. 👴

snarfed:
if you want to be lazy, you don't have to set up everything to build and run tests locally. you can just add the test and push to the PR branch, and [CircleCI should run the tests automatically](https://circleci.com/gh/snarfed/wordpress-micropub) and report results there.

lostfocus:
It was/is a good reason to get reacquainted with the whole system, but thanks. :)

snarfed:
this is a great contribution. good discussion on the use case up front, clean code, and a test. thank you so much! and congrats on your first merge here!

lostfocus:
Thank you very much. It was a delight.

PR #76, opened 2017-10-16
Fix Handling of Dates 

dshanske:
This addresses the issue of Micropub passing a publish date with a timezone that does not match the one set in WordPress. It also, since it has to parse this anyway, adds the timezone into the arguments passed to wp_insert_post so it can be actioned on filter, specifically for the purpose of a feature in my Simple Location plugin. It can override the displayed timezone, and if this is passed to the filter, I can set the property I need without having to do the calculations a second time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,9 @@ To set up PHPCodesniffer to test adherence to [WordPress Coding Standards](https
 #### 1.3 (unreleased)
 * Saves [access token response](https://tokens.indieauth.com/) in a post meta field `micropub_auth_response`.
 * Bug fix for `post_date_gmt`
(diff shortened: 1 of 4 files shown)


dshanske:
Unit tests now test for correct timezone handling by passing in a Pacific Timestamp and testing that the correct UTC/GMT stamp is what is generated.

snarfed:
thanks for the contribution! code looks good. i'll happily merge once there are both new and existing tests!

dshanske:
I've added a test for no timezone data.

snarfed:
looks great! thanks for the contribution, and for the tests!

PR #83, opened 2017-12-10
Fixes to Support Indiewebcamp Austin 2017 Hack Day Project

dshanske:
* Remove Own Your Swarm Checkin Override
* If Microformats2 enabled theme or Post Kinds, store generated content as a summary/excerpt not content...which is consistent with Microformats and Micropub
* Support nested bookmark, like, repost, etc properties (I intend to come back to this, but this will support what I'm doing for now)

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -520,25 +520,19 @@ class Micropub {
 		$props = static::$input['replace'] ?: static::$input['properties'];
 		$lines = array();
 
-		// Special case OwnYourSwarm's checkin property; auto generate content for
-		// it even if Post Kinds or an mf2-aware theme is installed. Discussion:
-		// https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-300805891
+		if ( $args['post_content'] &&
+			 ( current_theme_supports( 'microformats2' ) ||
+			   // Post Kinds: https://wordpress.org/plugins/indieweb-post-kinds/
+			   taxonomy_exists( 'kind' ) ) ) {
+			return $args;
+		}
+
 		$checkin = $props['checkin'][0];
 		if ( $checkin ) {
 			$name = $checkin['properties']['name'][0];
 			$urls = $checkin['properties']['url'];
 			$lines[] = '<p>Checked into <a class="h-card p-location" href="' .
 				($urls[1] ?: $urls[0]) . '">' . $name . '</a>.</p>';
-		}
-
-		if ( $args['post_content'] &&
-			 ( current_theme_supports( 'microformats2' ) ||
-			   // Post Kinds: https://wordpress.org/plugins/indieweb-post-kinds/
-			   taxonomy_exists( 'kind' ) ) ) {
-			if ( $lines ) {
-				$args['post_content'] .= "\n" . implode( "\n", $lines ) ;
-			}
-			return $args;
 		}
 
 		$verbs = array(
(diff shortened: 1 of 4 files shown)


snarfed:
thanks for the contribution! and sounds like IWC Austin was great, congrats!

in general, it's preferable to send different changes like this in different PRs. (you can do that with different branches on your fork repo.) not a big deal though.

can i get some context on removing the OYS special case? we added it for a reason...has something changed?

dshanske:
Post Kinds now supports check-ins.

dshanske:
Also supports nested properties, thus the other change.

dshanske:
I still want to knock off a bunch of issues after I cross a few other things off my list.

micropub.php
```diff
@@ -550,11 +544,16 @@ public static function generate_post_content( $args ) {
 		// interactions
 		foreach ( array_keys( $verbs ) as $prop ) {
 			$val = $props[ $prop ][0];
-			if ( $val ) {
+			if ( is_string( $val ) ) {
 				$lines[] = '<p>' . $verbs[ $prop ] .
 						   ' <a class="u-' . $prop . '" href="' .
 						   $val . '">' . $val . '</a>.</p>';
 			}
+			if ( is_array( $val ) && isset( $val['name'] ) && isset( $val['url'] ) ) {
```
snarfed: this mostly duplicates the last few lines. you'll want to unify them somehow.

micropub.php
```diff
@@ -570,8 +569,13 @@ public static function generate_post_content( $args ) {
 		// bookmark
 		if ( isset( $props['bookmark-of'] ) ) {
 			foreach ( $props['bookmark-of'] as $bookmark ) {
-				$lines[] = '<p>Bookmarked <a class="u-bookmark-of" href="' .
-						 $bookmark . '">' . $bookmark . '</a>.</p>';
+				if ( is_string( $bookmark ) ) {
+					$lines[] = '<p>Bookmarked <a class="u-bookmark-of" href="' .
+							 $bookmark . '">' . $bookmark . '</a>.</p>';
+				}
+				if ( is_array( $bookmark ) && isset( $bookmark['name'] ) && isset( $bookmark['url'] ) ) {
```
snarfed: same here.

tests/test_micropub.php
```diff
@@ -512,30 +512,6 @@ function test_create_checkin() {
 		$this->assertEquals( '<p>Checked into <a class="h-card p-location" href="http:/a/place">A Place</a>.</p>', $post->post_content );
 	}
 
-	function test_create_checkin_autogenerates_checkin_text_with_content_and_post_kinds() {
```
snarfed: you want to add tests, not remove them. :P we should definitely keep this test and at least use it to check the generated excerpt, if that's where this is instead of content.

snarfed (changes requested):
thanks again!

i'm open to all three changes here, but you have some work to do on them, and they definitely all need unit tests. you'll probably be happier if you split them up into separate PRs; it'll definitely be easier to iterate on them.

removing the OYS override still needs a decent amount of compatibility homework. i'm ok with removing it if the Post Kinds version that handles checkins is installed, but we need that version check, and we probably also want to keep generating it for microformats2 themes...unless you know something i don't? (and we ideally need tests for all those cases.)

otherwise, storing auto-generated content in excerpt and nested properties both generally sound good to me, with tests, and the comments here handled.

thanks again!

dshanske:
Based on the CS changes, I'm going to resubmit as requested as separate PRs once that is merged.

PR #29, opened 2016-07-06
Add Codesniffer file for WordPress Standards

dshanske:
This is the file @pfefferle set up and I use for checking against WordPress coding standards.

PR #86, opened 2017-12-13
Cleanup

dshanske:
This sets up a composer development configuration to do PHPCS testing for PHP compatibility and WordPress Coding Standards. It also fixes all Coding Standards issues with no new functionality.

Based on the output, the minimum PHP setting was declared as 5.3 as you use commands that are minimum 5.3.

Finally, a script was set up to generate the readme.md from the readme.txt so maintaining two files is easier.

All tests still pass. This just makes the code cleaner to work with.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -23,7 +23,10 @@ machine:
   environment:
     WP_CORE_DIR: $HOME/wordpress
     WP_TESTS_DIR: $HOME/wordpress-tests-lib
-
+    PATH: $HOME/.composer/vendor/bin:$PATH
 test:
+  pre:
+    - composer install
   override:
+    - ./vendor/bin/phpcs --config-set ignore_warnings_on_exit 1 --standard=phpcs.xml 
     - ./phpunit-5.7.phar --coverage-html $CIRCLE_ARTIFACTS
--- a/micropub.php
+++ b/micropub.php
@@ -1,35 +1,37 @@
 <?php
-/*
- Plugin Name: Micropub
- Plugin URI: https://github.com/snarfed/wordpress-micropub
- Description: <a href="https://indiewebcamp.com/micropub">Micropub</a> server.
- Protocol spec: <a href="https://www.w3.org/TR/micropub/">w3.org/TR/micropub</a>
- Author: Ryan Barrett
- Author URI: https://snarfed.org/
- Version: 1.3
-*/
+/**
+ * Plugin Name: Micropub
+ * Plugin URI: https://github.com/snarfed/wordpress-micropub
+ * Description: <a href="https://indiewebcamp.com/micropub">Micropub</a> server.
+ * Protocol spec: <a href="https://www.w3.org/TR/micropub/">w3.org/TR/micropub</a>
+ * Author: Ryan Barrett
+ * Author URI: https://snarfed.org/
+ * Text Domain: micropub
+ * Version: 1.3
+ */
 
-/* See README for supported filters and actions. */
+/* See README for supported filters and actions.
+ * Example command lines for testing:
+ * Form-encoded:
+ * curl -i -H 'Authorization: Bearer ...' -F h=entry -F name=foo -F content=bar \
+ *   -F photo=@gallery/snarfed.gif 'http://localhost/w/?micropub=endpoint'
+ * JSON:
+ * curl -v -d @body.json -H 'Content-Type: application/json' 'http://localhost/w/?micropub=endpoint'
+ *
+ * To generate an access token for testing:
+ * 1. Open this in a browser, filling in SITE:
+ *   https://indieauth.com/auth?me=SITE&scope=post&client_id=https://wordpress.org/plugins/micropub/&redirect_uri=https%3A%2F%2Findieauth.com%2Fsuccess
+ * 2. Log in.
+ * 3. Extract the code param from the URL.
+ * 4. Run this command line, filling in CODE and SITE (which logged into IndieAuth):
+ *   curl -i -d 'code=CODE&me=SITE&client_id=indieauth&redirect_uri=https://indieauth.com/success' 'https://tokens.indieauth.com/token'
+ * 5. Extract the access_token parameter from the response body.
+ */
 
-// Example command lines for testing:
-// Form-encoded:
-// curl -i -H 'Authorization: Bearer ...' -F h=entry -F name=foo -F content=bar \
-//   -F photo=@gallery/snarfed.gif 'http://localhost/w/?micropub=endpoint'
-// JSON:
-// curl -v -d @body.json -H 'Content-Type: application/json' 'http://localhost/w/?micropub=endpoint'
-//
-// To generate an access token for testing:
-// 1. Open this in a browser, filling in SITE:
-//   https://indieauth.com/auth?me=SITE&scope=post&client_id=https://wordpress.org/plugins/micropub/&redirect_uri=https%3A%2F%2Findieauth.com%2Fsuccess
-// 2. Log in.
-// 3. Extract the code param from the URL.
-// 4. Run this command line, filling in CODE and SITE (which logged into IndieAuth):
-//   curl -i -d 'code=CODE&me=SITE&client_id=indieauth&redirect_uri=https://indieauth.com/success' 'https://tokens.indieauth.com/token'
-// 5. Extract the access_token parameter from the response body.
-
-// For debugging purposes this will bypass Micropub authentication
-// in favor of WordPress authentication
-// Using this to test querying(q=) parameters quickly
+/* For debugging purposes this will bypass Micropub authentication
+ * in favor of WordPress authentication
+ * Using this to test querying(q=) parameters quickly
+ */
 if ( ! defined( 'MICROPUB_LOCAL_AUTH' ) ) {
 	define( 'MICROPUB_LOCAL_AUTH', '0' );
 }
@@ -47,19 +49,16 @@ if ( ! defined( 'MICROPUB_DRAFT_MODE' ) ) {
 	define( 'MICROPUB_DRAFT_MODE', '0' );
 }
 
-if ( ! class_exists( 'Micropub' ) ) :
-
-add_action( 'init', array( 'Micropub', 'init' ) );
+add_action( 'init', array( 'Micropub_Plugin', 'init' ) );
 
 /**
  * Micropub Plugin Class
  */
-class Micropub {
+class Micropub_Plugin {
 	// associative array
 	public static $request_headers;
 
-    // associative array, read from JSON or form-encoded input. populated by
-    // load_input().
+	// associative array, read from JSON or form-encoded input. populated by load_input().
 	protected static $input;
 
 	// associative array, populated by authorize().
@@ -110,8 +109,10 @@ class Micropub {
 
 		static::load_input();
 		if ( WP_DEBUG ) {
-			error_log( 'Micropub Data: ' . serialize( $_GET ) . ' ' .
-					   serialize( static::$input ) );
+			error_log(
+				'Micropub Data: ' . serialize( $_GET ) . ' ' .
+					serialize( static::$input )
+			);
 		}
 		static::$input = apply_filters( 'before_micropub', static::$input );
 
@@ -125,9 +126,9 @@ class Micropub {
 			$user_id = static::authorize();
 		}
 
-		if ( $_SERVER['REQUEST_METHOD'] == 'GET' && isset( $_GET['q'] ) ) {
+		if ( 'GET' === $_SERVER['REQUEST_METHOD'] && isset( $_GET['q'] ) ) {
 			static::query_handler( $user_id );
-		} elseif ( $_SERVER['REQUEST_METHOD'] == 'POST' ) {
+		} elseif ( 'POST' === $_SERVER['REQUEST_METHOD'] ) {
 			self::post_handler( $user_id );
 		} else {
 			static::error( 400, 'Unknown Micropub request' );
@@ -142,28 +143,33 @@ class Micropub {
 	 */
 	private static function authorize() {
 		// find the access token
-		$auth = static::get_header( 'authorization' );
+		$auth  = static::get_header( 'authorization' );
 		$token = $_POST['access_token'];
-		if ( ! $auth && ! $token) {
+		if ( ! $auth && ! $token ) {
 			static::handle_authorize_error( 401, 'missing access token' );
 		}
 
 		$resp = wp_remote_get(
-			MICROPUB_TOKEN_ENDPOINT, array( 'headers' => array(
-				'Content-Type' => 'application/x-www-form-urlencoded',
-				'Authorization' => $auth ?: 'Bearer ' . $token,
-			) ) );
+			MICROPUB_TOKEN_ENDPOINT, array(
+				'headers' => array(
+					'Content-Type'  => 'application/x-www-form-urlencoded',
+					'Authorization' => $auth ?: 'Bearer ' . $token,
+				),
+			)
+		);
 		$code = wp_remote_retrieve_response_code( $resp );
 		$body = wp_remote_retrieve_body( $resp );
 		parse_str( $body, $params );
 		$scopes = explode( ' ', $params['scope'] );
 
-		if ( (int) ( $code / 100 ) != 2 ) {
+		if ( (int) ( $code / 100 ) !== 2 ) {
 			static::handle_authorize_error(
-				$code, 'invalid access token: ' . $body );
-		} elseif ( ! in_array( 'post',  $scopes ) && ! in_array( 'create',  $scopes ) ) {
+				$code, 'invalid access token: ' . $body
+			);
+		} elseif ( ! in_array( 'post', $scopes, true ) && ! in_array( 'create', $scopes, true ) ) {
 			static::handle_authorize_error(
-				401, 'access token is missing post or create scope; got `' . $params['scope'] . '`' );
+				401, 'access token is missing post or create scope; got `' . $params['scope'] . '`'
+			);
 		}
 
 		parse_str( $body, $resp );
@@ -171,26 +177,48 @@ class Micropub {
 
 		static::$micropub_auth_response = $resp;
 
-		// look for a user with the same url as the token's `me` value. search both
-		// with and without trailing slash.
-		foreach ( array_merge( get_users( array( 'search' => $me ) ),
-							   get_users( array( 'search' => $me . '/' ) ) )
-				  as $user ) {
-			if ( untrailingslashit( $user->user_url ) == $me ) {
-				return $user->ID;
-			}
+		// look for a user with the same url as the token's `me` value.
+		$user = static::user_url( $me );
+		if ( $user ) {
+			return $user;
 		}
 
 		// no user with that url. if the token is for this site itself, allow it and
 		// post as the default user
 		$home = untrailingslashit( home_url() );
-		if ( $home != $me ) {
+		if ( $home !== $me ) {
 			static::handle_authorize_error(
-				401, 'access token URL ' . $me . " doesn't match site " . $home . ' or any user' );
+				401, 'access token URL ' . $me . " doesn't match site " . $home . ' or any user'
+			);
 		}
 
-		return NULL;
+		return null;
 	}
+
+	/**
+	 * Try to match a user with a URL with or without trailing slash.
+	 *
+	 * @param string $me URL to match
+	 *
+	 * @return null|int Return user ID if matched or null
+	**/
+	public static function user_url( $me ) {
+		$search = array(
+			'search'         => $me,
+			'search_columns' => array( 'url' ),
+		);
+		$users  = get_users( $search );
+
+		$search['search'] = $me . '/';
+		$users            = array_merge( $users, get_users( $search ) );
+		foreach ( $users as $user ) {
+			if ( untrailingslashit( $user->user_url ) === $me ) {
+				return $user->ID;
+			}
+		}
+		return null;
+	}
+
 
 	/**
 	 * Parse the micropub request and render the document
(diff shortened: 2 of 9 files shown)


readme.md
```diff
@@ -1,8 +1,20 @@
-# wordpress-micropub [![Circle CI](https://circleci.com/gh/snarfed/wordpress-micropub.svg?style=svg)](https://circleci.com/gh/snarfed/wordpress-micropub)
```
snarfed: this should be preserved. if it can go into readme.txt and look right in the wordpress plugin directory, i'm all for that!

readme.md
```diff
@@ -95,13 +127,11 @@ To set up your local environment to run the unit tests:
 1. Open `wordpress-tests-lib/wp-tests-config.php` and add a slash to the end of the ABSPATH value. No clue why it leaves off the slash; it doesn't work without it.
 1. Run `phpunit` in the repo root dir. If you set `WP_CORE_DIR` and `WP_TESTS_DIR` above, you'll need to set them for this too. You should see output like this:
 
-    ```
```
snarfed: github's markdown renderer is pretty finicky and wants this exact syntax - four space indent, blank line before, and three backticks - to render as a code block _inside_ the list item. you'll probably need to preserve that exactly for it to still render right.

readme.md
```diff
@@ -62,28 +73,49 @@ Alternatively, you can set `MICROPUB_LOCAL_AUTH` to 1 to use WordPress's interna
...
+
+If that doesn't work either, you may need to ask your hosting provider to whitelist the `Authorization` header for your account. If they refuse, you can [pass it through Apache with an alternate name](https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-299569822), but [you'll need to edit this plugin's code to read from that alternate name](https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-299569822).
+
 
-If that doesn't work, you may need to ask your hosting provider to whitelist the `Authorization` header for your account. If they refuse, you can [pass it through Apache with an alternate name](https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-299569822), but [you'll need to edit this plugin's code to read from that alternate name](https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-299569822).
+## Upgrade Notice 
 
+None yet.
 
-### License
```
snarfed: this section needs to be preserved.

readme.md
```diff
@@ -11,15 +23,15 @@ Once you've installed and activated the plugin, try using [Quill](http://quill.p
...
+## WordPress details 
 
-#### Filters and hooks
 
+### Filters and hooks 
 Adds two filters:
 
 `before_micropub( $input )`
 
-Called before handling a Micropub request. Returns `$input`, possibly modified.
```
snarfed: odd that the backticks here got lost. any idea why?

readme.md
```diff
@@ -1,8 +1,20 @@
...
+**Tags:** micropub, publish  
+**Requires at least:** 4.4  
+**Requires PHP:** 5.3  
+**Tested up to:** 4.9.1  
+**Stable tag:** trunk  
+**License:** CC0  
+**License URI:** http://creativecommons.org/publicdomain/zero/1.0/  
+**Donate link:** -  
 
-A [Micropub](http://micropub.net/) server plugin for [WordPress](https://wordpress.org/). Available in the WordPress plugin directory at [wordpress.org/plugins/micropub](https://wordpress.org/plugins/micropub/).
```
snarfed: ...and this longer version of the sentence should ideally be preserved.

readme.md
```diff
@@ -1,8 +1,20 @@
-# wordpress-micropub [![Circle CI](https://circleci.com/gh/snarfed/wordpress-micropub.svg?style=svg)](https://circleci.com/gh/snarfed/wordpress-micropub)
+# Micropub
```
snarfed: ...and this section should definitely be removed for github.

micropub.php
```diff
@@ -49,389 +51,419 @@
 
 if ( ! class_exists( 'Micropub' ) ) :
```
snarfed: looks like this change indents the entire file due to this outer condition. we definitely don't want that. :P if wordpress standards say to, i'm ok with disobeying that. can you configure PHPCS to skip this indent?

micropub.php
```diff
@@ -49,389 +51,419 @@
...
-					$args[ $name ] = NULL;
+			}
+
+			// replace
+			$replace = static::$input['replace'];
+			if ( $replace ) {
+				if ( ! is_array( $replace ) ) {
+					static::error( 400, 'replace must be an object' );
 				}
-				if ( in_array( 'category', $delete ) ) {
```
snarfed: this code change seems suspicious. was it unrelated and accidentally got lumped into this PR?

micropub.php
```diff
@@ -49,389 +51,419 @@
...
+			$args              = static::store_mf2(
+				static::store_geodata(
+					static::generate_post_content( $args )
+				)
+			);
+			if ( WP_DEBUG ) {
+				error_log( 'wp_update_post with args: ' . serialize( $args ) );
+			}
 
-		static::default_file_handler( $post_id );
```
snarfed: this one too. do you maybe need to rebase on master? it'll be easier to tell one way or the other after the global indent is reverted.

snarfed (changes requested):
hey, thanks! maybe not high priority, but i know you've been interested in this stuff for a while: #20, #29, etc. :P

a few thoughts:

* i'd _love_ to auto generate one readme from the other, but i'm skeptical. i've spent a bit of time trying to do this before, and github's and the wordpress plugin directory's markdown flavors differ a bit. each one has things that seem to always get lost converting in either direction. if you can get it working without losing anything significant, i'm all for it!
* i'm not that familiar with composer. can you add some basic commands for setting it up and using it to the readme? also, looks like composer.lock was auto generated. should we let developers do that locally, instead of committing it?
* if we like PHPCS (or whatever auto-formats the code), should we try to run that automatically, either on git pre-commit or in circle? happy to defer that to a later PR if you want.
* let's skip the global indent change in micropub.php. details in comments there.

micropub.php
dshanske: It isn't deleted. It is just lower down on line 425.

micropub.php
dshanske: According to WordPress coding standards, it shouldn't be skipped. It appears in a specific example in Best Practices. 

https://developer.wordpress.org/plugins/the-basics/best-practices/

readme.md
dshanske: I'll do a PR to reconcile the old and new readmes. But the goal was that both be versions of the same file. They effectively were, but the conversion was copied by hand.

readme.md
dshanske: No, the conversion is done in software. But let me have a look at the converter.

readme.md
dshanske: This means, again, it wasn't in both files. I'm going to do a piece by piece comparison and edit the readme and rerun the converter to include all.

dshanske:
The reason I wanted to finally get this one addressed is, the likelihood is we'll figure out the rendering issue when we get to the scheduled meeting, one way or the other. Having the plugin comply with WordPress standards is just a clerical task that, once done, is easy to maintain. 

I'm going to read through the Github documentation guidelines and figure this out. Looking at most Github repos that are WordPress plugins, they usually have the markdown and text versions match(complete with the WordPress header). I'll see what I can learn.

Specifically with composer, the .lock file notes the development versions and most people state it should be committed to ensure any development is consistent with the versions used. Admittedly, the development files in this case are PHPCS, the WordPress coding standards, and the PHP Compatibility check, so versioning may not be so important.

As for composer, @pfefferle introduced Grunt to most of the other Indieweb plugins and I adopted that for some as well to run several tasks at once. Micropub didn't need two. Composer is common to PHP projects. It brings in dependencies and can run scripts. So, the only command needed is... composer install or composer update. All the dependencies are development ones only, compared to Semantic Linkbacks which uses composer to bring in the latest version of php-mf2 and the Emoji detection library.

What would you think about putting composer and such notes into a CONTRIBUTING document? Or other separate document? To address the issue with the READMEs?

PHPCS has an autofix script called PHPCBF. It will automatically fix simple issues.That could be programmed to run by Circle-CI, or PHPCS could, with a failure on error, requiring a manual fix. I believe we have the others configured to fail on PHPCS error, not try to auto-fix if possible.

Either way, after I solve this simpler issue, the next one I will likely tackle is the multiple properties issue, which, if not addressed, will cause your code, when paired with Post Kinds, to put in lines like, "Bookmarked Array"

snarfed:
> I'm going to read through the Github documentation guidelines and figure this out. Looking at most Github repos that are WordPress plugins, they usually have the markdown and text versions match(complete with the WordPress header). I'll see what I can learn.

nice! that kind of research is a great idea. i'd still probably choose to omit the wordpress header from readme.md, and keep the existing license section, but i'm definitely glad to hear other precedent.

> Specifically with composer, the .lock file notes the development versions and most people state it should be committed to ensure any development is consistent with the versions used. Admittedly, the development files in this case are PHPCS, the WordPress coding standards, and the PHP Compatibility check, so versioning may not be so important.

ok!

> Composer is common to PHP projects. It brings in dependencies and can run scripts. So, the only command needed is... composer install or composer update. ...What would you think about putting composer and such notes into a CONTRIBUTING document? Or other separate document? To address the issue with the READMEs?

great! it's not enough to warrant a separate file, and it would fit nicely in the existing _Development_ section in the readmes.

> Either way, after I solve this simpler issue, the next one I will likely tackle is the multiple properties issue, which, if not addressed, will cause your code, when paired with Post Kinds, to put in lines like, "Bookmarked Array"

ah, great! definitely sounds like a bug worth fixing.

micropub.php
snarfed: understood, but i'm going to choose to disobey their standards in this case.

readme.md
snarfed: understood! it's a worthy goal. let's reconcile them in this PR, so we know it's possible with the automated script, then i'll be on board.

dshanske:
Made an attempt to address all comments

dshanske:
How do these changes look so far?

The outstanding issues appear to be..

1. The WordPress header in the readme.md, which if it is an issue, I can add the scripting to store the header separately and generate both files, I suppose.
1. The check for class isn't necessary if you prefix your class, so I did, and that fixed the alignment. Also a good move as Micropub, while descriptive, could be reused.

micropub.php
```diff
@@ -173,23 +179,26 @@ private static function authorize() {
...
 		// look for a user with the same url as the token's `me` value. search both
 		// with and without trailing slash.
-		foreach ( array_merge( get_users( array( 'search' => $me ) ),
-							   get_users( array( 'search' => $me . '/' ) ) )
-				  as $user ) {
-			if ( untrailingslashit( $user->user_url ) == $me ) {
+		foreach ( array_merge(
+			get_users( array( 'search' => $me ) ),
+			get_users( array( 'search' => $me . '/' ) )
+		)
```
snarfed: oh man this kind of indentation - an inner paren all the way back at the top level - is so problematic and misleading when reading code! any way we can loosen PHPCS a bit with indentation rules like this?

micropub.php
```diff
@@ -142,28 +143,33 @@ public static function parse_query( $wp ) {
 	 */
 	private static function authorize() {
 		// find the access token
-		$auth = static::get_header( 'authorization' );
+		$auth  = static::get_header( 'authorization' );
```
snarfed: ugh. this variable init aligning is annoying when there are so few vars, like just two here. ugh.

micropub.php
```diff
@@ -200,34 +209,34 @@ private static function authorize() {
...
-		$unknown = array_diff( $synd_requested, $synd_supported );
+		$unknown        = array_diff( $synd_requested, $synd_supported );
 
 		if ( $unknown ) {
-			static::error( 400, 'Unknown syndicate-to targets: ' . implode(', ', $unknown ) );
+			static::error( 400, 'Unknown syndicate-to targets: ' . implode( ', ', $unknown ) );
 
-		// create
-		} elseif ( ! $url || $action == 'create' ) {
+			// create
```
snarfed: this is wrong. same with the rest of the comment indents below. it should either be where it was before, or if that's impossible, append it to the line below.

micropub.php
```diff
@@ -369,7 +390,7 @@ private static function update() {
 				static::error( 400, 'replace must be an object' );
 			}
 			foreach ( static::mp_to_wp( array( 'properties' => $replace ) )
-					  as $name => $val ) {
+				as $name => $val ) {
```
snarfed: this seems wrong. it should be in two spaces to show that it's inside the paren above, right?

micropub.php
```diff
@@ -383,32 +404,39 @@ private static function update() {
...
+					$args['post_category'] = array_diff(
+						$args['post_category'] ?: array(),
+						$delete_args['post_category']
+					);
 				}
 			} elseif ( is_array( $delete ) ) {
 				foreach ( static::mp_to_wp( array( 'properties' => array_flip( $delete ) ) )
-						  as $name => $_ ) {
-					$args[ $name ] = NULL;
+					as $name => $_ ) {
```
snarfed: ditto

micropub.php
```diff
@@ -383,32 +404,39 @@ private static function update() {
...
 			}
 		}
 
-		// tell wordpress to preserve published date explicitly, otherwise
+		// tell WordPress to preserve published date explicitly, otherwise
 		// wp_update_post sets it to the current time
 		$args['edit_date'] = true;
-		$args = static::store_mf2( static::store_geodata(
-			static::generate_post_content( $args ) ) );
+		$args              = static::store_mf2(
```
snarfed: so ugly :(

micropub.php
```diff
@@ -525,25 +556,25 @@ public static function generate_post_content( $args ) {
 		// https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-300805891
 		$checkin = $props['checkin'][0];
 		if ( $checkin ) {
-			$name = $checkin['properties']['name'][0];
-			$urls = $checkin['properties']['url'];
+			$name    = $checkin['properties']['name'][0];
+			$urls    = $checkin['properties']['url'];
 			$lines[] = '<p>Checked into <a class="h-card p-location" href="' .
-				($urls[1] ?: $urls[0]) . '">' . $name . '</a>.</p>';
+			( $urls[1] ?: $urls[0] ) . '">' . $name . '</a>.</p>';
```
snarfed: wow this is *very* misleading. implies that it's a standalone statement, but it's really part of the line above. we need to loosen PHPCS to prevent changes like this.

micropub.php
```diff
@@ -525,25 +556,25 @@ public static function generate_post_content( $args ) {
...
 		}
 
 		if ( $args['post_content'] &&
-			 ( current_theme_supports( 'microformats2' ) ||
-			   // Post Kinds: https://wordpress.org/plugins/indieweb-post-kinds/
-			   taxonomy_exists( 'kind' ) ) ) {
+		( current_theme_supports( 'microformats2' ) ||
```
snarfed: ditto

micropub.php
```diff
@@ -552,26 +583,26 @@ public static function generate_post_content( $args ) {
...
-						   $val . '">' . $val . '</a>.</p>';
+					' <a class="u-' . $prop . '" href="' .
+					$val . '">' . $val . '</a>.</p>';
 			}
 		}
 
 		if ( isset( $props['rsvp'] ) ) {
 			$lines[] = '<p>RSVPs <data class="p-rsvp" value="' . $props['rsvp'][0] .
-				'">' . $props['rsvp'][0] . '</data>.</p>';
+			'">' . $props['rsvp'][0] . '</data>.</p>';
```
snarfed: ditto

snarfed (changes requested):
your work to unify the readmes looks great! thank you! only a couple minor nits there. and the instructions for composer look good too, thanks for adding them.

>     The check for class isn't necessary if you prefix your class, so I did, and that fixed the alignment. Also a good move as Micropub, while descriptive, could be reused.

clever idea! nice workaround. `Micropub_Plugin` is a pretty awkward name, but no matter, i'll take it. :P

> The WordPress header in the readme.md, which if it is an issue, I can add the scripting to store the header separately and generate both files, I suppose.

you can easily remove it with e.g. `tail`. something like (untested) `tail -n +12 readme.txt | wp2md > readme.md`.

other outstanding things:

* i get the desire to match PHPCS and wordpress's standards strictly, but some of the code style changes really hurt readability. hopefully we can find a way to loosen PHPCS a bit?
* git pre-commit trigger? for both PHPCS/PHPBF and auto-generating readme.md?

circle.yml
```diff
@@ -23,7 +23,10 @@ machine:
   environment:
     WP_CORE_DIR: $HOME/wordpress
     WP_TESTS_DIR: $HOME/wordpress-tests-lib
-
+    PATH: $HOME/.composer/vendor/bin:$PATH
 test:
+  pre:
+    - composer install
   override:
+    - ./vendor/bin/phpcs --config-set ignore_warnings_on_exit 1 --standard=phpcs.xml
```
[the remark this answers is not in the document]
dshanske: It just tests it. I'd rather the changes be done by manual commit, partially due the issus you raised.

micropub.php
dshanske: I hated this line for different reasons beforehand. This is where I wanted to change it back to having these functions act as filters, but I think the three layers of functions in 1 entry is confusing in the first place. For now, I think I will convert them to 3 separate lines for readability.

micropub.php
dshanske: I was able to get some of these reversed. It looks like, when it realigned the whole page, it removed it. I'm manually adjusting and it isn't flagging.

circle.yml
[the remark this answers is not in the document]
dshanske: By the way, the main plugin class for Webmentions is Webmention_Plugin.

dshanske:
I think that covers everything.

readme.md
```diff
@@ -1,25 +1,33 @@
-# wordpress-micropub [![Circle CI](https://circleci.com/gh/snarfed/wordpress-micropub.svg?style=svg)](https://circleci.com/gh/snarfed/wordpress-micropub)
+A [Micropub](http://micropub.net/) server plugin.
 
-A [Micropub](http://micropub.net/) server plugin for [WordPress](https://wordpress.org/). Available in the WordPress plugin directory at [wordpress.org/plugins/micropub](https://wordpress.org/plugins/micropub/).
```
snarfed: this still needs to be preserved.

micropub.php
snarfed: friendly ping

micropub.php
```diff
@@ -369,7 +386,7 @@ private static function update() {
 				static::error( 400, 'replace must be an object' );
 			}
 			foreach ( static::mp_to_wp( array( 'properties' => $replace ) )
-					  as $name => $val ) {
+				  as $name => $val ) {
```
snarfed: thanks for fixing this!

micropub.php
snarfed: ok!

micropub.php
snarfed: perfect. thank you!

micropub.php
```diff
@@ -552,26 +581,26 @@ public static function generate_post_content( $args ) {
...
 
 		if ( isset( $props['rsvp'] ) ) {
 			$lines[] = '<p>RSVPs <data class="p-rsvp" value="' . $props['rsvp'][0] .
-				'">' . $props['rsvp'][0] . '</data>.</p>';
+			  '">' . $props['rsvp'][0] . '</data>.</p>';
 		}
 
 		// event
-		if ( static::$input['type'] == array( 'h-event' ) ) {
+		if ( array( 'h-event' ) === static::$input['type'] ) {
```
snarfed: it seems so silly that wordpress wants both yoda order _and_ `===`. i get that yoda order prevents accidental assignment, but that's not possible with `===`. meh.

micropub.php
```diff
@@ -863,12 +908,14 @@ public static function get_mf2( $post_id ) {
 	}
 
 	public static function error( $code, $message ) {
-		static::respond( $code, array(
-			'error' => ($code == 403) ? 'forbidden' :
-					   ($code == 401) ? 'insufficient_scope' :
-					   'invalid_request',
-			'error_description' => $message,
-		) );
+		static::respond(
```
snarfed: wow. ugly. :P

snarfed (commented):
getting closer!

my main question is around the manual code style tweaks you've made for my comments. you said some of them seemed ok, and phpcs didn't complain. is that true for all of them? it'd be nice if we don't have to keep cleaning up after phpcs continually.

otherwise almost there!

micropub.php
dshanske: That first one, the get_users...the sniffer is pretty specific about not letting me adjust that.

dshanske:
Trying again. Doing the best possible.

micropub.php
snarfed: that's unfortunate, since it's pretty misleading. can you configure it to be more relaxed? often you can disable individual rules in code beautifiers.

i like using phpcs, but it's just a dumb tool, like all tools. i don't want us to be a slave to it when we know better. :P

snarfed:
thanks for sticking with this!

nudge on this q:

> my main question is around the manual code style tweaks you've made for my comments. you said some of them seemed ok, and phpcs didn't complain. is that true for all of them? it'd be nice if we don't have to keep cleaning up after phpcs continually.

also, once we figure out that last style nit, could you please squash the commits?

snarfed:
oh, and could you also please test this manually a bit? it's encouraging that the tests pass, but the code changes are big enough and invasive enough that i'd also like to see some manual testing, with multiple clients and code paths. (or webmention.rocks!)

dshanske:
I actually did. Posted from Quill onto my test site. I will squash the commits now

dshanske:
Squashed now.

readme.md
snarfed: ping?

micropub.php
snarfed: ping

dshanske:
I tried to address the formatting issue by refactoring the url search into its own function. Similar to the other issue, having three things done in a single line, even if split up into pieces, can affect readability. So if that is the goal, having it in steps solves the display problem and makes it easier to read. If this works, I'll squash again.

dshanske:
Also, in a future PR, should really unit test the new function. But want to get this done first.

micropub.php
```diff
@@ -177,20 +177,12 @@ private static function authorize() {
...
-			if ( untrailingslashit( $user->user_url ) === $me ) {
-				return $user->ID;
-			}
+		// look for a user with the same url as the token's `me` value.
+		$user = static::user_url( $me );
+		if ( $user ) {
+			return $user;
 		}
 
-		// no user with that url. if the token is for this site itself, allow it and
```
snarfed: please restore :P

snarfed:
thanks again for sticking with this!

my only remaining question is, ping on this:

> my main question is around the manual code style tweaks you've made for my comments. you said some of them seemed ok, and phpcs didn't complain. is that true for all of them? it'd be nice if we don't have to keep cleaning up after phpcs continually.

if phpcs leaves the current code alone, and you squash the new commits, i'm happy to merge this!

dshanske:
It won't change its mind unless the standards change.

dshanske:
All merged and fixed now.

snarfed:
woo! thanks again.

snarfed (approved):

PR #88, opened 2017-12-19
Fix variable based errors

dshanske:
This cleans up a few related issues from the error log. Namely.

* If there is no access token, avoids throwing an error by initializing the variable to null
* The query was being checked off of _GET directly, not off of the filter. There is no reason why filtering shouldn't be available for queries.
* $args isn't used on a query, so setting the parameter to null to avoid another error.

Shouldn't be anything that controversial here, I just hate filling my error log.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -126,7 +126,7 @@ class Micropub_Plugin {
 			$user_id = static::authorize();
 		}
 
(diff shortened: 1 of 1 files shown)


snarfed:
this is great! very much appreciated. thank you!

snarfed:
this is great! very much appreciated. thank you!

PR #89, opened 2017-12-21
Support nested properties in content generation

dshanske:
This avoids nested properties from producing phrases like Bookedmarked Array. It adds tests to ensure same. Currently, like the previous version, it does not support multiple URLs in a property for content generation.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -596,15 +596,35 @@ class Micropub_Plugin {
 			'like-of'     => 'Likes',
 			'repost-of'   => 'Reposted',
 			'in-reply-to' => 'In reply to',
+			'bookmark-of' => 'Bookmarked',
 		);
 
 		// interactions
 		foreach ( array_keys( $verbs ) as $prop ) {
-			$val = $props[ $prop ][0];
+			if ( wp_is_numeric_array( $props[ $prop ] ) ) {
+				$val = $props[ $prop ][0];
+			} else {
+				$val = $props[ $prop ];
+			}
 			if ( $val ) {
-				$lines[] = '<p>' . $verbs[ $prop ] .
-				  ' <a class="u-' . $prop . '" href="' .
-				  $val . '">' . $val . '</a>.</p>';
+				// Supports nested properties by turning single value properties into nested
+				// https://micropub.net/draft/#nested-microformats-objects
+				if ( is_string( $val ) ) {
+					$val = array(
+						'url' => $val,
+					);
+				}
+				if ( ! isset( $val['name'] ) && isset( $val['url'] ) ) {
+					$val['name'] = $val['url'];
+				}
+				if ( isset( $val['url'] ) ) {
+					$lines[] = sprintf(
+						'<p>%1s <a class="u-%2s" href="%3s">%4s</a>.</p>',
+						$verbs[ $prop ],
+						$prop, $val['url'],
+						$val['name']
+					);
+				}
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


snarfed:
woo! thank you for the fix, and for the tests! i'm definitely looking forward to seeing this fixed.

i assume this is https://micropub.net/draft/#nested-microformats-objects ? which clients generate nested objects? quill?

dshanske:
Right now, Post Kinds replaces the URL in a Quill bookmark post with a nested name + url using the before_micropub filter if the property is only a URL. This has made the response end up as Bookmarked Array. This will address that.

Either way, nested properties are a thing, and several Micropub clients support them, though not Quill.

dshanske:
https://github.com/dshanske/indieweb-post-kinds/issues/137

micropub.php
```diff
@@ -593,15 +593,31 @@ public static function generate_post_content( $args ) {
...
+				$val = $props[ $prop ];
+			}
 			if ( $val ) {
-				$lines[] = '<p>' . $verbs[ $prop ] .
-				  ' <a class="u-' . $prop . '" href="' .
-				  $val . '">' . $val . '</a>.</p>';
+				$l = '<p>' . $verbs[ $prop ] .
+				' <a class="u-' . $prop . '" href="';
+				if ( is_string( $val ) ) {
+					$l .= $val . '">' . $val;
```
snarfed: the closing `>` should probably be added outside all of these conditionals. otherwise, if `$val` is something other than a string or array, or if `$val['url']` isn't set, it won't get added at all right now.

(speaking of which, that would be a good test to add. :P)

...or actually, something like `sprintf` would probably be a lot more readable and understandable than all these conditionals. mind trying that instead?

micropub.php
```diff
@@ -593,15 +593,31 @@ public static function generate_post_content( $args ) {
...
+				}
+				if ( is_array( $val ) ) {
```
snarfed: mind adding a comment with a link to the nested properties part of the spec here? https://micropub.net/draft/#nested-microformats-objects

tests/test_micropub.php
```diff
@@ -908,6 +908,41 @@ function test_create_bookmark() {
 , $post->post_content );
 	}
 
+	function test_create_nested_bookmark() {
```
snarfed: mind adding a comment that this example markup below is generated by Post Kinds? and ideally a link to details in the Post Kind docs.

snarfed (changes requested):
great first pass! thanks again.

micropub.php
```diff
@@ -593,15 +593,31 @@ public static function generate_post_content( $args ) {
...
+			if ( wp_is_numeric_array( $props[ $prop ] ) ) {
+				$val = $props[ $prop ][0];
+			} else {
+				$val = $props[ $prop ];
+			}
 			if ( $val ) {
-				$lines[] = '<p>' . $verbs[ $prop ] .
-				  ' <a class="u-' . $prop . '" href="' .
-				  $val . '">' . $val . '</a>.</p>';
+				$l = '<p>' . $verbs[ $prop ] .
```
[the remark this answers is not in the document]
dshanske: $lines[] is the array of lines. $l is one line which is merged as $lines[] lower down to cover the cases. I can rename it $line. Would that work?

micropub.php
dshanske: Happy to do a sprintf. Would definitely make things a bit easier.

tests/test_micropub.php
dshanske: Will note it appropriately. But it could be easily done by a client as within spec.

dshanske:
How's this?

micropub.php
```diff
@@ -596,15 +596,36 @@ public static function generate_post_content( $args ) {
 			'like-of'     => 'Likes',
 			'repost-of'   => 'Reposted',
 			'in-reply-to' => 'In reply to',
+			'bookmark-of' => 'Bookmarked',
 		);
 
 		// interactions
+		$line = '<p>%1s <a class="u-%2s" href="%3s">%4s</a>.</p>';
```
snarfed: minor nit: please inline this into the `sprintf` call for clarity.

PR #90, opened 2017-12-21
Add filter for custom query responses

dshanske:
I'd like to be able to allow for custom queries in the event another plugin wants to extend the supported properties and/or add new ones in future.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,7 +21,7 @@ This project is placed in the public domain. You may also use it under the [CC0 
 
 
 ### Filters and hooks 
-Adds two filters:
+Adds three filters:
 
 `before_micropub( $input )`
 
@@ -30,6 +30,10 @@ Called before handling a Micropub request. Returns `$input`, possibly modified.
 `micropub_syndicate-to( $synd_urls, $user_id )`
 
 Called to generate the list of `syndicate-to` targets to return in response to a query. Returns `$synd_urls`, an array, possibly modified.
+
+`micropub_query( $resp, $input )`
+
+$resp defaults to null. If the return value is non-null, it should be an associative array that is encoded as JSON and will be returned in place of the normal micropub response.
 
 ...and one hook:
 
@@ -150,6 +154,7 @@ into markdown and saved to readme.md.
 * Correctly handle published times that are in a different timezone than the site.
 * Set minimum version to PHP 5.3
 * Adhere to WordPress Coding Standards
+* Add `micropub_query` filter
 
 
 ### 1.2 (2017-06-25) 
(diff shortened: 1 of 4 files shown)


snarfed:
sure! we definitely need to all agree on the full coordinating logic we've been talking about in https://github.com/indieweb/wordpress-uf2/issues/30, and not do it piecemail with just individual plugins...but this is good to have anyway. thanks!

it definitely needs a test though. mind adding that?

snarfed:
...actually, `before_micropub` already fires on queries. we should just make that handle this, instead of adding a new filter, right?

dshanske:
Before_micropub filters the input. If we filtered the input, how would that affect the output?

dshanske:
Re the coordinating logic, that was the logic for the mf2 stuff. The query response I've been thinking of would be an experimental one. Querying for venue info. It's on the brainstorming page. But expanding the existing query responses would be useful if I wrote a media endpoint plugin that was an optional extension. I have no plans to do a media endpoint right now though.

readme.md
```diff
@@ -31,6 +31,10 @@ Called before handling a Micropub request. Returns `$input`, possibly modified.
 
 Called to generate the list of `syndicate-to` targets to return in response to a query. Returns `$synd_urls`, an array, possibly modified.
 
+`pre_micropub_query( $resp, $input )`
+
+$resp defaults to null. Used to add custom query response if return is not null
```
snarfed: minor nit: the second sentence could be clearer. how about:

> $resp defaults to null. If the return value is non-null, it should be an associative array that is encoded as JSON and returned in place of the normal micropub response.

snarfed (changes requested):
thanks for the explanations! sounds good. let's name it `micropub_query`, no `pre_`. and we need a unit test. :P after that, i'm ready to merge!

dshanske:
Unit test and recommended fixes deployed.

tests/test_micropub.php
```diff
@@ -256,6 +256,43 @@ function syndicate_to() {
...
+		$this->check( 200, $expected );
+
+		$this->assertEquals( $_GET, static::$before_micropub_input );
+		$this->assertEquals( $_GET, static::$after_micropub_input );
+		$this->assertNull( static::$after_micropub_args );
+	}
+
+
+	function test_custom_query_still_returns_default() {
+		function custom_default_query( $return, $input ) {
```
snarfed: this should be a class method and resused in both tests instead of duplicated.

alternatively, feel free to just move lines 286-288 into the previous test, and remove this one.

PR #91, opened 2017-12-22
Add post status support and tests

dshanske:
I think the title says it all. Unit tests included.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -82,6 +82,18 @@ class Micropub_Plugin {
 		add_action( 'send_headers', array( $cls, 'http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
+		// Register Setting
+		register_setting(
+			'writing', // Option Group Writing
+			'micropub_default_post_status', // Setting Name
+			array(
+				'type'              => 'string',
+				'description'       => 'Default Post Status for Micropub Server',
+				'sanitize_callback' => 'sanitize_text_field',
+				'show_in_rest'      => false,
+				'default'           => MICROPUB_DRAFT_MODE ? 'draft' : 'publish', // If not set can be set with deprecated property
+			)
+		);
 	}
 
 	/**
@@ -349,7 +361,10 @@ class Micropub_Plugin {
 		if ( $user_id ) {
 			$args['post_author'] = $user_id;
 		}
-		$args['post_status'] = MICROPUB_DRAFT_MODE ? 'draft' : 'publish';
+		$args['post_status'] = static::post_status( static::$input );
+		if ( ! $args['post_status'] ) {
+			static::error( 400, 'Invalid Post Status' );
+		}
 		if ( WP_DEBUG ) {
 			error_log( 'wp_insert_post with args: ' . serialize( $args ) );
 		}
@@ -479,6 +494,39 @@ class Micropub_Plugin {
 			return;
 		}
 		static::respond( $code, $msg );
+	}
+
+	private static function post_status( $mf2 ) {
+		$props          = $mf2['properties'];
+		$visibilitylist = array( array( 'private' ), array( 'public' ) );
+		if ( isset( $props['visibility'] ) ) {
+			if ( ! in_array( $props['visibility'], $visibilitylist, true ) ) {
+				// Returning null will cause the server to return a 400 error
+				return null;
+			}
+			if ( array( 'private' ) === $props['visibility'] ) {
+				return 'private';
+			}
+		}
+		if ( ! isset( $props['post-status'] ) ) {
+			return get_option( 'micropub_default_post_status' );
+		} else {
+			//  According to the proposed specification these are the only two properties supported.
+			// https://indieweb.org/Micropub-extensions#Post_Status
+			// For now these are the only two we will support even though WordPress defaults to 8 and allows custom
+			// But makes it easy to change
+			$statuslist = array( array( 'published' ), array( 'draft' ) );
+			if ( ! in_array( $props['post-status'], $statuslist, true ) ) {
+				// Returning null will cause the server to return a 400 error
+				return null;
+			}
+			// Map published to the WordPress property publish.
+			if ( array( 'published' ) === $props['post-status'] ) {
+				return 'publish';
+			}
+			return 'draft';
+		}
+		// Execution will never reach here
 	}
 
 	/**
--- a/readme.md
+++ b/readme.md
@@ -53,6 +53,15 @@ Stores [microformats2](http://microformats.org/wiki/microformats2) properties in
 
 Does *not* support multithreading. PHP doesn't really either, so it generally won't matter, but just for the record.
 
+Supports Experimental Properties for [Post Status](https://indieweb.org/Micropub-extensions#Post_Status) and [Visibility](https://indieweb.org/Micropub-extensions#Visibility). Visibility can be either
+`public` or `private`. Setting it to `private` will set the post status to private. Post Status may be set to either `published` or `draft`. If visibility or post status are not set to one of these 
+options, the plugin will return HTTP 400 with body:
+
+    {
+      "error": "invalid_request",
+      "error_description": "Invalid Post Status"
+    }
+
 WordPress has a [whitelist of file extensions that it allows in uploads](https://codex.wordpress.org/Uploading_Files#About_Uploading_Files_on_Dashboard). If you upload a file in a Micropub extension that doesn't have an allowed extension, the plugin will return HTTP 400 with body:
 
     {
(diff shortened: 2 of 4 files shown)


micropub.php
```diff
@@ -516,6 +515,25 @@ private static function mp_to_wp( $mf2 ) {
...
+			if ( ! in_array( $props['post-status'], $statuslist, true ) ) {
+				$args['post_status'] = apply_filters( 'micropub_default_post_status', 'publish' );
+			}
+			// Map published to the WordPress property publish.
+			if ( 'published' === $props['post-status'] ) {
+				$args['post_status'] = 'publish';
+			}
+		}
+		if ( isset( $props['visibility'] ) && ( array( 'private' ) === $props['visibility'] ) ) {
+				$args['post_status'] = 'private';
```
snarfed: minor nit: indented too much

micropub.php
```diff
@@ -516,6 +526,25 @@ private static function mp_to_wp( $mf2 ) {
...
+		if ( ! isset( $props['post-status'] ) ) {
+			$args['post_status'] = get_option( 'micropub_default_post_status' );
+		} else {
+			//  According to the proposed specification these are the only two properties supported.
+			// https://indieweb.org/Micropub-extensions#Post_Status
+			// For now these are the only two we will support even though WordPress defaults to 8 and allows custom
+			// But makes it easy to change
+			$statuslist = array( array( 'published' ), array( 'draft' ) );
+			if ( ! in_array( $props['post-status'], $statuslist, true ) ) {
+				$args['post_status'] = get_option( 'micropub_default_post_status' );
```
snarfed: we should return a 400 error here, or whatever the micropub spec says to do, instead of silently ignoring the status they requested.

dshanske: As I sort of alluded to in chat, https://www.w3.org/TR/micropub/#unrecognized-properties

"If the request includes properties that the server does not recognize, it MUST ignore unrecognized properties and create the post with the values that are recognized."

So, if the server doesn't recognize the status, it should ignore it by this statement. Since post-status is a proposed extension to Micropub, and not part of the specification...

https://indieweb.org/Micropub-extensions#Post_Status

dshanske: Thinking it out, I get the point here.

dshanske:
I rewrote it to reject unknown properties. Also took the liberty of adding a note to the readme as this property is experimental.

micropub.php
```diff
@@ -481,6 +496,37 @@ private static function handle_authorize_error( $code, $msg ) {
...
 
+	private static function post_status( $mf2 ) {
+		$props = $mf2['properties'];
+		if ( isset( $props['visibility'] ) ) {
+			if ( array( 'private' ) === $props['visibility'] ) {
+				return 'private';
+			}
+			// Public or Private are the only options
+			if ( array( 'public' ) !== $props['visiblity'] ) {
+				return null;
```
snarfed: this should be an error too.

micropub.php
```diff
@@ -481,6 +496,37 @@ private static function handle_authorize_error( $code, $msg ) {
...
+			// But makes it easy to change
+			$statuslist = array( array( 'published' ), array( 'draft' ) );
+			if ( ! in_array( $props['post-status'], $statuslist, true ) ) {
+				return null;
+			}
+			// Map published to the WordPress property publish.
+			if ( 'published' === $props['post-status'] ) {
+				return 'publish';
+			}
+			return $props['post-status'];
```
snarfed: this returns an array, right? probably should be just `return 'draft';`? sounds like we need another unit test. :P

micropub.php
```diff
@@ -349,7 +361,10 @@ private static function create( $user_id ) {
 		if ( $user_id ) {
 			$args['post_author'] = $user_id;
 		}
-		$args['post_status'] = MICROPUB_DRAFT_MODE ? 'draft' : 'publish';
```
snarfed: i think we actually need to keep the constant around for backward compatibility, and fall back to it if the micropub request doesn't set post status and the setting is unset. sorry for not mentioning that sooner!

dshanske: I can make the default of the option triggered by it.

micropub.php
dshanske: Returning null causes the 400 error to be triggered.

micropub.php
dshanske: I have one for the creation of draft status. It seems to work. What would you suggest the other unit test does?

dshanske:
I've changed it so MICROPUB_DRAFT_MODE will set the default for the option if the option is not set. That preserves backward compatibility.

micropub.php
snarfed: ah. great! that's worth a comment. also `visiblity` above is misspelled, which means we need a unit test for this error handling code path. :P

micropub.php
snarfed: ah, that's true, thanks! that's a good enough test. i still don't understand this code, though. `$props['post-status']` is always an array here, right? so the conditional on line 522 will always be false, and this line will always return that array...so then when we set `$args['post_status']` to an array instead of a string on line 364, and then call `wp_update_post()`, wordpress handles the array value ok?

readme.txt
```diff
@@ -153,7 +161,7 @@ into markdown and saved to readme.md.
 * Adhere to WordPress Coding Standards
 * Add `micropub_query` filter
 * Support Nested Properties in Content Generation 
-
+* Remove `MICROPUB_DRAFT_MODE` configuration option in favor of setting option
```
snarfed: s/Remove/Deprecate/, please

readme.md
```diff
@@ -156,7 +164,7 @@ into markdown and saved to readme.md.
 * Adhere to WordPress Coding Standards
 * Add `micropub_query` filter
 * Support Nested Properties in Content Generation 
-
+* Remove `MICROPUB_DRAFT_MODE` configuration option in favor of setting option
```
snarfed: s/Remove/Deprecate/, please

snarfed (changes requested):

dshanske:
Remove replaced with deprecate and yes, it handles arrays where it doesn't expect them.

snarfed:
> yes, it handles arrays where it doesn't expect them.

ok! that means line 522 (post status `published`) is broken, then, since it's an array, which probably means we need another unit test for that. :P

dshanske:
I fully agree. I didn't test publish and should have because published->publish. And when I checked it...low and behold...issue.  Fixed, squashed.

micropub.php
```diff
@@ -481,6 +496,37 @@ private static function handle_authorize_error( $code, $msg ) {
...
+			if ( ! in_array( $props['post-status'], $statuslist, true ) ) {
+				return null;
+			}
+			// Map published to the WordPress property publish.
+			if ( array( 'published' ) === $props['post-status'] ) {
+				return 'publish';
+			}
+			return $props['post-status'];
+		}
+		return null;
```
snarfed: minor nit: probably remove this, for readability, since it will never actually be executed.

snarfed: ...minor nit: that also means you could drop the `} else {` above and move its block up to top level.

micropub.php
snarfed: ping!

micropub.php
```diff
@@ -481,6 +496,37 @@ private static function handle_authorize_error( $code, $msg ) {
...
+			// But makes it easy to change
+			$statuslist = array( array( 'published' ), array( 'draft' ) );
+			if ( ! in_array( $props['post-status'], $statuslist, true ) ) {
+				return null;
+			}
+			// Map published to the WordPress property publish.
+			if ( array( 'published' ) === $props['post-status'] ) {
+				return 'publish';
+			}
+			return $props['post-status'];
```
snarfed: minor nit: probably change this to `return 'draft';` for consistency and readability, since we return `'publish'` as a string above, not an array, and for readability since it's guaranteed to be `'draft'` here.

snarfed (commented):
thanks for the `published` fix!

micropub.php
dshanske: I have a test in the unit tests for someone sending in an unrecognized property. Won't it actually be executed if someone sends in an unrecognized error. Returning null will, when it is called return the unrecognized value error

micropub.php
dshanske: Again, unless we support additional post-status properties in future. Is it better to make it so to add properties, we just add them to the whitelist above, or wait till we need it?

dshanske:
Fixed the issue again.

micropub.php
snarfed: good point! predicting the future is difficult and fraught, though, so i lean toward only designing for the present, which is a known quantity. let's go with `return 'draft';`.

micropub.php
snarfed: nope, this line will never run. the `if` and `else` paths above both end in `return`s. for clarity, let's replace it with an `// execution will never reach here` comment.

micropub.php
snarfed: ping on commenting on each `return null` in this function that they result in an HTTP 400.

dshanske:
Requested changes made.

micropub.php
snarfed: thanks for the fixes! this request keeps falling through the cracks though. once these two comments are in, i'll happily merge!

dshanske:
What am I missing here still?

snarfed:
@dshanske https://github.com/snarfed/wordpress-micropub/pull/91#discussion-diff-158431609

image

dshanske:
Those comments are in thoi

snarfed:
oh you're right! sorry about that, evidently the diff i was looking at in that section was an old snapshot. yay, this looks great!

PR #92, opened 2017-12-23
Store Generated Content in Excerpt when Support for Microformats2 or Post Kind is Declared

dshanske:
This is the other part of the previously closed PR, although I've made some changes to the previous proposal. 

Firstly, the content is generated in 100% of cases, and is stored in the post_excerpt/summary field, unless a summary was already generated by the client. 

This is consistent with the client-side example of  https://www.w3.org/TR/micropub/#unrecognized-properties.

Post Kinds certainly supports rendering, and a theme that declares support for Microformats2 should support the scenario where excerpt is set but content is not, as both are accepted properties.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -568,35 +568,22 @@ class Micropub_Plugin {
 	 * and friends.
 	 */
 	public static function generate_post_content( $args ) {
-		$props = static::$input['replace'] ?: static::$input['properties'];
-		$lines = array();
-
-		// Special case OwnYourSwarm's checkin property; auto generate content for
-		// it even if Post Kinds or an mf2-aware theme is installed. Discussion:
-		// https://github.com/snarfed/wordpress-micropub/issues/56#issuecomment-300805891
-		$checkin = $props['checkin'][0];
-		if ( $checkin ) {
-			$name    = $checkin['properties']['name'][0];
-			$urls    = $checkin['properties']['url'];
-			$lines[] = '<p>Checked into <a class="h-card p-location" href="' .
-				( $urls[1] ?: $urls[0] ) . '">' . $name . '</a>.</p>';
-		}
-
-		if ( $args['post_content'] &&
-			  ( current_theme_supports( 'microformats2' ) ||
-			  // Post Kinds: https://wordpress.org/plugins/indieweb-post-kinds/
-			  taxonomy_exists( 'kind' ) ) ) {
-			if ( $lines ) {
-				$args['post_content'] .= "\n" . implode( "\n", $lines );
-			}
-			return $args;
-		}
-
-		$verbs = array(
+		$props      = static::$input['replace'] ?: static::$input['properties'];
+		$lines      = array();
+		$kindneeded = array();
+		$verbs      = array(
 			'like-of'     => 'Likes',
 			'repost-of'   => 'Reposted',
 			'in-reply-to' => 'In reply to',
 			'bookmark-of' => 'Bookmarked',
+		);
+		// Kinds are the term used for Indieweb post types in the Post Kinds plugin as post types are a WordPress term already.
+		// This endeavors to map properties to their kinds to determine if Post Kinds supports it.
+		$kindmap = array(
+			'like-of'     => 'like',
+			'repost-of'   => 'repost',
+			'in-reply-to' => 'reply',
+			'bookmark-of' => 'bookmark',
 		);
 
 		// interactions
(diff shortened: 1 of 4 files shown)


dshanske:
I'll even go submit some PRs for the themes that don't support this now.

pfefferle:
Why the `post_excerpt` instead of `post_content`?

dshanske:
Because it is a summary. Most of the Micropub clients send a text version in summary in the event that the server cannot render the mf2 properties.

pfefferle:
But you can‘t compare a checkin summary with a post excerpt, can you? I would create a plain text post to be compatible with themes!

dshanske:
If you don't recognize the checkin property, then you would render as a note, in theory. Either way, this might be addressed if we ever solve the mf2 conflict.

snarfed:
this looks great! thank you!

my only request is that we check the post kinds version, and limit this new behaior to just post kinds with at least the version that handles checkins, so we maintain compatibility with other versions. and likewise, we don't do it for `microformats2` themes until we know they support this too.

dshanske:
There isn't an easy way to guarantee getting the plugin version that I see. You can get one for any plugin if you know where it is installed, but getting into that is complicated. So I tried a different tactic. Post Kinds stores the kinds the user sets as active in an option called kinds_termlist. So, I'm mapping all properties that micropub recognizes to the names of the kinds('reply', 'bookmark', 'checkin', etc) and then if that isn't a supported kind, then it will save the content.

dshanske:
I've also removed the microformats2 override. To your point, none of the microformats2 themes actually supports rendering those properties now. So why are we exempting them?

micropub.php
```diff
@@ -568,36 +568,21 @@ private static function mp_to_wp( $mf2 ) {
...
+		$props      = static::$input['replace'] ?: static::$input['properties'];
+		$lines      = array();
+		$kindneeded = '';
+		$verbs      = array(
 			'like-of'     => 'Likes',
 			'repost-of'   => 'Reposted',
 			'in-reply-to' => 'In reply to',
 			'bookmark-of' => 'Bookmarked',
 		);
+		$kindmap    = array(
```
snarfed: minor nit: please comment that the "kinds" here are for the Post Kinds plugin.

micropub.php
```diff
@@ -625,17 +610,39 @@ public static function generate_post_content( $args ) {
...
-			$lines[] = static::generate_event( static::$input );
+			$lines[]    = static::generate_event( static::$input );
+			$kindneeded = 'event';
+		}
+		if ( isset( $kindmap[ $kindneeded ] ) ) {
+			$kindneeded = $kindmap[ $kindneeded ];
+		}
+
+		if ( ! $args['post_excerpt'] &&
+			  // Post Kinds: https://wordpress.org/plugins/indieweb-post-kinds/
```
snarfed: minor nit: this needs a bit more elaboration, e.g. that Post Kinds sets `kind_termslist` to an array of string kinds that the user has enabled.

snarfed (changes requested):
> I'm mapping all properties that micropub recognizes to the names of the kinds('reply', 'bookmark', 'checkin', etc) and then if that isn't a supported kind, then it will save the content.

great idea! i like it. thank you!

one catch: a post can have multiple mf2 props to render, so `$kindneeded` needs to be an array, not a single kind string, and this function needs to render and return all mf2 props that post kinds won't handle (or aren't currently enabled).

> I've also removed the microformats2 override. To your point, none of the microformats2 themes actually supports rendering those properties now. So why are we exempting them?

great point! is that true? you've researched them all and confirmed? if so, agreed, dropping it sgtm. please note all of the themes you researched in the changelog.

micropub.php
dshanske: I have done as you ask.

dshanske:
I hope I covered everything.

micropub.php
```diff
@@ -625,17 +612,43 @@ public static function generate_post_content( $args ) {
...
+			$lines[]      = static::generate_event( static::$input );
+			$kindneeded[] = 'event';
+		}
+		foreach ( $kindneeded as $key => $value ) {
+			if ( isset( $kindmap[ $value ] ) ) {
+				$kindneeded[ $key ] = $kindmap[ $value ];
+			}
+		}
+		// A list of enabled kinds is stored by the Post Kind plugin in kinds_termslist. This compares used properties to supported kinds.
+		$kindneeded = array_intersect( $kindneeded, get_option( 'kind_termslist', array() ) );
```
snarfed: you actually want `array_diff` instead of `array_intersect`, right? since you want to find out which values in `$kindneeded` are _not_ in `kind_termslist` and render those, right?

snarfed: ...for example, if `kind_termslist` is `like, repost, reply` and `kindneeded` is `rsvp, reply`, that means post kinds won't render the rsvp part of the current post, so we still need to, right? ie we should use `array_diff` to narrow down to just the kinds that we need render manually, and render them, right?

this PR is starting to feel more complicated than it's worth. am i overthinking it?

snarfed: (at minimum, we probably need a new unit test for when the current post and `kind_termslist` have different but overlapping mf2 properties.)

dshanske: Well, Post Kinds doesn't support multiple types in the same post. You can't reply and bookmark in the same post. Admittedly, this is an artificial construct. I will fix the array issue and a unit test. But, for now, this seems a simple compromise.

snarfed: understood. Post Kinds may not support multiple types per post, but _micropub_ does, which is why we still need `array_diff`. :P

...but also, as i mentioned earlier: we do need all these changes for this PR to be correct, but it's starting to feel too complicated. i'm feeling less and less confident about this change overall, and this approach to compatibility with Post Kinds. i'd love to hear other ideas!

dshanske: I could go back to the drawing board, I suppoe. There was the other option, which was to turn generate_post_content back into a filter. Then Post Kinds could turn it off itself.

dshanske:
I'm going to try the alternate method, I think.

snarfed:
thanks! I'm guessing it's similar to the very first version of this PR? i may be more open to it now that we've seen the alternative. :P

snarfed:
tentatively closing since we merged #94, the alternative to this. feel free to reopen!

PR #94, opened 2017-12-30
Add Post Generation Filter

dshanske:
This tries a different tactic. It adds a filter. Anything that wants to remove post content generation can therefore do so.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -82,6 +82,10 @@ class Micropub_Plugin {
 		add_action( 'send_headers', array( $cls, 'http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
+
+		// Post Content Filter
+		add_filter( 'micropub_post_content', array( $cls, 'generate_post_content' ), 1, 2 );
+
 		// Register Setting
 		register_setting(
 			'writing', // Option Group Writing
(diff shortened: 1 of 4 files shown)


micropub.php
```diff
@@ -82,6 +82,10 @@ public static function init() {
 		add_action( 'send_headers', array( $cls, 'http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
+
+		// Post Content Filter
+		add_filter( 'micropub_post_content', array( $cls, 'generate_post_content' ), 10, 2 );
```
snarfed: we should probably use a much higher priority than 10 here, since we always want `generate_post_content` to run first, right? maybe 1 or 0?

micropub.php
```diff
@@ -469,9 +477,13 @@ private static function update() {
 		// Store metadata from Microformats Properties
 		$args = static::store_mf2( $args );
 		$args = static::store_geodata( $args );
-		// Generate Post Content
-		$args = static::generate_post_content( $args );
 
+		// Generate Post Content
+		$post_content = isset( $args['post_content'] ) ? $args['post_content'] : '';
```
snarfed: unfortunate that this block is duplicated both here and in `create`. maybe not quite enough to warrant unifying them though. meh.

snarfed (changes requested):
this is great! apologies for making you try out the alternative approach first. sometimes you don't know the right path until you walk down the wrong one a ways. :P

i only have a few minor requests. otherwise ready to merge!

micropub.php
dshanske: 10 is default. But why should this run before adding properties?

micropub.php
dshanske: I thought of that

micropub.php
snarfed: so they can see the content this plugin generates! they may want to tweak it instead of replacing it entirely.

dshanske: Good point, but in that case, I'll set the other options to 11 and 12. It would be atypical otherwise.

snarfed: no, i think we want 1 or 0. remember, _anyone_ can use a filter. you can't control other people's plugins and themes. you want to make the normal, default path work for other people, without them doing anything special. :P

dshanske:
As requested, but there is something about the test I'm missing

micropub.php
```diff
@@ -676,6 +669,14 @@ public static function generate_post_content( $args ) {
 			}
 		}
 
+		$checkin = $props['checkin'][0];
```
snarfed: minor nit: should this be `isset( $props['checkin'] )` instead? it was there before, i know you didn't write it, but just checking since we're here.

snarfed (commented):
looks pretty straightforward. the test is failing because it's missing the `</p>`: https://circleci.com/gh/snarfed/wordpress-micropub/188#action-18

you're set up to run tests locally, right? always a good idea to do that, and fix any failures, before pushing.

micropub.php
dshanske: I do run tests locally. I pushed it because I just couldn't see it.

dshanske:
If this meets with your approval and you merge it, can you push 1.3 with these changes?

snarfed:
looks great! thanks again for sticking with this. merging now; i'll release 1.3 soon.

Issue #99, opened 2018-01-15
Multiple HTTP Headers are Being Overwritten


dshanske:
https://secure.php.net/manual/en/function.header.php

Code uses the header function...which unless replace is set to false, will not allow multiple headers of the same type.

snarfed:
thanks again!

PR #100, opened 2018-01-15
Fix header related issues

dshanske:
Discovered in working on Indieauth tests that the Micropub plugin wasn't actually sending all of the headers because the replace parameter wasn't set to false. So it was always using the html parameter without anyone noticing up until now.

I'm also going to as part of this, move the firing of the plugin initialization to an earlier hook which fires before the setup_theme hooks, which it should with this type of plugin.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -49,7 +49,7 @@ if ( ! defined( 'MICROPUB_DRAFT_MODE' ) ) {
 	define( 'MICROPUB_DRAFT_MODE', '0' );
 }
 
-add_action( 'init', array( 'Micropub_Plugin', 'init' ) );
+add_action( 'plugins_loaded', array( 'Micropub_Plugin', 'init' ) );
 
 /**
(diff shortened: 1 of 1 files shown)


snarfed:
[background from IRC](https://chat.indieweb.org/wordpress/2018-01-15#t1516029078557900):

> snarfed: morning GWG! re https://github.com/snarfed/wordpress-micropub/issues/99 ...odd, i don't actually see headers get replaced with the current code. e.g. it currently happily outputs multiple Link headers right now
>   your PR looks fine, i'm just curious
> [pfefferle]: me neither
>   but it is only for the same rel
>   that means, if you have multiple authentication provider for example
>   I had the issue when using multiple “hub” header for websub
>   perhaps GWG found out, while working on the IndieAuth plugin…
> snarfed: ah ok

snarfed:
thanks! merging.

PR #101, opened 2018-01-16
Improve Header Generation

dshanske:
* Separate micropub from indieauth headers so can be removed/adjusted if need be separately
* Allow for using the config options that the Indieauth plugin uses. If not set, default to the existing definitions


snarfed:
lgtm! please update the _Authentication and authorization_ and _Configuration Options_ sections in the docs to mention that the indieauth plugin overrides the constants if it's installed.

dshanske:
Will do.

dshanske:
Redid the readme per request.

Issue #96, opened 2018-01-01
IndieAuth Plugin and Micropub Plugin


dshanske:
Since I'm working on IndieAuth support in the IndieAuth Plugin, invariably, we might end up with a conflict as I iterate. Wanted to open an issue to discuss possibilities.

https://github.com/indieweb/wordpress-indieauth/

1. Right now, Micropub preempts WordPress if the micropub query variable is set and launches into a completely different set of code, so it will not be affected by the code I have hooking into determine_current_user which duplicates the same functionality. 
1. It uses 'indieauth_token_endpoint' and 'indieauth_authorization_endpoint' as option names. Should be easy to look there and fall back on the declaration if they aren't set. 

I'm not sure if it is worth making it so Micropub uses the Indieauth plugin version of the code over the built in one.

snarfed:
thanks for discussing ahead of time! all this analysis makes sense and sounds good to me.

snarfed:
btw this looks similar to #85? let's maybe close one or the other.

dshanske:
That is about mf2 whereas this is about Indieauth

snarfed:
oops, sorry, i mistyped. i meant #84.

dshanske:
It is, but now it has been done

snarfed:
great! closed.

PR #103, opened 2018-01-18
Update MICROPUB_LOCAL_AUTH config option

dshanske:
I'm expanding the LOCAL_AUTH option to disable built-in auth. That means it can be used in both the original noted way or a new way...allowing the Indieauth plugin to handle authentication. Since my hope is to get a built-in endpoint in the plugin someday, this will support that.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -78,15 +78,18 @@ class Micropub_Plugin {
 		add_action( 'parse_query', array( $cls, 'parse_query' ) );
 
 		// endpoint discovery
-		add_action( 'wp_head', array( $cls, 'indieauth_html_header' ), 99 );
 		add_action( 'wp_head', array( $cls, 'micropub_html_header' ), 99 );
-		add_action( 'send_headers', array( $cls, 'indieauth_http_header' ) );
 		add_action( 'send_headers', array( $cls, 'micropub_http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'micropub_jrd_links' ) );
-		add_filter( 'host_meta', array( $cls, 'indieauth_jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'micropub_jrd_links' ) );
-		add_filter( 'webfinger_user_data', array( $cls, 'indieauth_jrd_links' ) );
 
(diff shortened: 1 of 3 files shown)


dshanske:
This should address #96 if merged.

micropub.php
```diff
@@ -136,11 +139,8 @@ public static function parse_query( $wp ) {
 		}
 		static::$input = apply_filters( 'before_micropub', static::$input );
 
-		// For debug purposes be able to bypass Micropub auth with WordPress auth
+		// Be able to bypass Micropub auth with other auth
 		if ( MICROPUB_LOCAL_AUTH ) {
-			if ( ! is_user_logged_in() ) {
-				auth_redirect();
```
snarfed: is this unnecessary? why?

snarfed:
cool! if you've manually tested and this still uses wordpress auth if indieauth isn't installed, and correctly uses indieauth if it's installed, then i'll happily merge. (pending the one question.) thanks!

dshanske:
It does because Indieauth hooks into WordPress auth while Micropub hooks in earlier

micropub.php
dshanske: The auth redirect never worked properly.

snarfed: oh! um, ok. do we need it? i guess micropub requests generally aren't interactive, so this isn't useful anyway...? do we currently return HTTP 401 when `MICROPUB_LOCAL_AUTH` is set but the user isn't logged in?

snarfed:
> It does because Indieauth hooks into WordPress auth while Micropub hooks in earlier

apologies, just to make sure i understand...you mean you have manually tested this with and without the indieauth plugin installed? and both cases work as expected?

dshanske:
Yes.

micropub.php
dshanske: That I will check

dshanske:
I added in the 401 failure message. It is returned by the Indieauth code, but Micropub doesn't use WP_Error, so it doesn't get picked up. I may try to adjust that.

dshanske:
Specifically, start using WP_Error

snarfed:
looks good, thanks!

Issue #97, opened 2018-01-14
No Action or URL


dshanske:
Got this when testing some modifications to posting notes from Quill. Action and URL are not required properties on all objects. Going to see about fixing this at some point, but wanted to make note.

2018/01/13 20:16:25 [error] 19211#19211: *37441 FastCGI sent in stderr: "PHP message: Micropub Data: a:2:{s:8:"micropub";s:8:"endpoint";s:1:"q";s:6:"config";} a:2:{s:8:"micropub";s:8:"endpoint";s:1:"q";s:6:"config";}" while reading response header from upstream, client: 173.230.155.197, server: tiny.n9n.us, request: "GET /?micropub=endpoint&q=config HTTP/1.1", upstream: "fastcgi://127.0.0.1:9000", host: "tiny.n9n.us"
2018/01/13 20:16:55 [error] 19212#19212: *37445 FastCGI sent in stderr: "PHP message: Micropub Data: a:1:{s:8:"micropub";s:8:"endpoint";} a:2:{s:4:"type";a:1:{i:0;s:7:"h-entry";}s:10:"properties";a:1:{s:7:"content";a:1:{i:0;s:18:"Testing something.";}}}
PHP message: PHP Notice:  Undefined index: action in wp-content/plugins/wordpress-micropub/micropub.php on line 250
PHP message: PHP Notice:  Undefined index: url in wp-content/plugins/wordpress-micropub/micropub.php on line 251
PHP message: PHP Notice:  Undefined index: syndicate-to in wp-content/plugins/wordpress-micropub/micropub.php on line 255
PHP message: PHP Warning:  array_diff(): Argument #1 is not an array in wp-content/plugins/wordpress-micropub/micropub.php on line 256
PHP message: PHP Notice:  Undefined index: slug in wp-content/plugins/wordpress-micropub/micropub.php on line 571
PHP message: PHP Notice:  Undefined index: name in wp-content/plugins/wordpress-micropub/micropub.php on line 571
PHP message: PHP Notice:  Undefined index: summary in wp-content/plugins/wordpress-micropub/micropub.php on line 571
PHP message: PHP Notice:  Undefined index: replace in wp-content/plugins/wordpress-micropub/micropub.php on line 637
PHP message: PHP Notice:  Undefined index: like-of in wp-content/plugins/wordpress-micropub/micropub.php on line 649
[...]

PR #108, opened 2018-01-19
Bump Minimum Version Requirements and Initialize Variables

dshanske:
This should address #105 and #97 in regards to the initialization of variables and ensuring that the WordPress features in the plugin align with the minimum version they were implemented in.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -259,7 +259,7 @@ class Micropub_Plugin {
 
 		// check that we support all requested syndication targets
 		$synd_supported = apply_filters( 'micropub_syndicate-to', array(), $user_id );
-		$synd_requested = static::$input['properties']['syndicate-to'];
+		$synd_requested = isset( static::$input['properties']['syndicate-to'] ) ? static::$input['properties']['syndicate-to'] : array();
 		$unknown        = array_diff( $synd_requested, $synd_supported );
 
 		if ( $unknown ) {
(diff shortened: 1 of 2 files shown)


micropub.php
```diff
@@ -254,12 +254,12 @@ public static function user_url( $me ) {
 	 */
 	public static function post_handler( $user_id ) {
 		$status = 200;
-		$action = static::$input['action'];
-		$url    = static::$input['url'];
+		$action = isset( static::$input['action'] ) ? static::$input['action'] : 'create';
```
snarfed: we have so many of these, i wonder if we should start using something like the `get()` helper here: http://heap.ch/blog/2015/12/19/php-default-values/#convenience . not necessary for this PR, just an idea. thoughts?

snarfed (approved):
lgtm, thanks! merging.

micropub.php
dshanske: I have a similar function I created in most of my plugins called ifset that looks pretty much the same.

snarfed: ah, nice i'm all for adding it here and using it everywhere! (and i like `get` better than `isset`, since it returns a value and not a boolean...but that's a bikeshed. :P)

also, ooh, php 7 added an `??` operator for this! https://wiki.php.net/rfc/isset_ternary . nice to know.

dshanske: WordPress will hopefully get there before I retire.

PR #110, opened 2018-01-20
Replace isset with get

dshanske:
Ask and ye shall receive

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -107,6 +107,13 @@ class Micropub_Plugin {
 		);
 	}
 
+	public static function get( $array, $key, $default = array() ) {
+		if ( is_array( $array ) ) {
+			return isset( $array[ $key ] ) ? $array[ $key ] : $default;
+		}
+		return $default;
+	}
+
 	/**
 	 * Adds some query vars
 	 *
@@ -133,8 +140,8 @@ class Micropub_Plugin {
 		static::load_input();
 		if ( WP_DEBUG ) {
 			error_log(
-				'Micropub Data: ' . serialize( $_GET ) . ' ' .
-					serialize( static::$input )
+				'Micropub Data: ' . wp_json_encode( $_GET ) . ' ' .
+					wp_json_encode( static::$input )
 			);
 		}
 		static::$input = apply_filters( 'before_micropub', static::$input );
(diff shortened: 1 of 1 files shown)


snarfed:
ahhh this is so good! thank you!!! mind switching all of the existing `isset`s to it? not urgent.

i'm also tempted to move it out of the class, put i guess that puts it in the global PHP namespace, which is a bad idea...? yeah probably never mind. :P

snarfed:
friendly ping! this would be great to get in.

dshanske:
We can put it into the global namespace if we prefix it

snarfed:
eh, i'm fine with it as a static method, unless you feel strongly otherwise.

dshanske:
@snarfed I cleaned out a bunch of get related errors with undefined indexes and changed the signature of the get function to be more array index friendly. 

Also found the same variable name was used in two different contexts by accident, so changed it.

I did make one change that isn't strictly in the PR notification. I changed serialized in the error_log to json_encode because I found it easier to read. I think I originally put the serialize in.

micropub.php
```diff
@@ -925,28 +932,27 @@ public static function store_micropub_auth_response( $args ) {
 	 */
 	public static function store_mf2( $args ) {
 		$props = static::$input['properties'];
-		if ( $props ) {
-			if ( ! isset( $args['meta_input'] ) ) {
-				$args['meta_input'] = array();
-			}
-			$type = static::$input['type'];
+		if ( ! isset( $args['ID'] ) && $props ) {
```
snarfed: why was `! isset( $args['ID'] )` added here? seems unrelated?

snarfed (changes requested):
hey, thanks! lgtm other than one errant change.

micropub.php
dshanske: Because before, it would run the remaining lines, which don't apply on creation, generating an error.

dshanske: When I went to turn it into get, I realized that it was trying to run the replacement code on create.

snarfed: you mean it was running the create code on replace? (and delete and add?) i think that was ok as is, but this is fine too. ok!

snarfed (approved):

PR #111, opened 2018-01-21
Disable if Indieauth

dshanske:
This will disable if the class indicating the Indieauth Plugin exists.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -84,7 +84,7 @@ class Micropub_Plugin {
 		add_filter( 'webfinger_user_data', array( $cls, 'micropub_jrd_links' ) );
 
 		// Disable adding headers if local auth is set
-		if ( ! MICROPUB_LOCAL_AUTH ) {
+		if ( MICROPUB_LOCAL_AUTH  || ! class_exists( 'IndieAuth_Plugin' ) ) {
 			add_action( 'wp_head', array( $cls, 'indieauth_html_header' ), 99 );
 			add_action( 'send_headers', array( $cls, 'indieauth_http_header' ) );
 			add_filter( 'host_meta', array( $cls, 'indieauth_jrd_links' ) );
@@ -140,7 +140,7 @@ class Micropub_Plugin {
 		static::$input = apply_filters( 'before_micropub', static::$input );
 
 		// Be able to bypass Micropub auth with other auth
-		if ( MICROPUB_LOCAL_AUTH ) {
+		if ( MICROPUB_LOCAL_AUTH || class_exists( 'IndieAuth_Plugin' ) ) {
 			$user_id = get_current_user_id();
 			if ( ! $user_id ) {
 				static::handle_authorize_error( 401, 'Unauthorized' );
--- a/readme.md
+++ b/readme.md
@@ -81,7 +81,8 @@ Supports the full OAuth2/IndieAuth authentication and authorization flow. Defaul
 
 If the token's `me` value matches a WordPress user's URL, that user will be used. Otherwise, the token must match the site's URL, and no user will be used.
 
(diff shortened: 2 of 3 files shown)


readme.md
```diff
@@ -81,7 +81,8 @@ Supports the full OAuth2/IndieAuth authentication and authorization flow. Defaul
 
 If the token's `me` value matches a WordPress user's URL, that user will be used. Otherwise, the token must match the site's URL, and no user will be used.
 
-Alternatively, you can set `MICROPUB_LOCAL_AUTH` to 1 to disable the plugin's authorization function, for example if you want authorization to be done by WordPress or another plugin such as the Indieauth plugin.
+Alternatively, you can set `MICROPUB_LOCAL_AUTH` to 1 to disable the plugin's authorization function, for example if you want authorization to be done by WordPress or another plugin. It will also 
+disable if the IndieAuth plugin is installed.
```
snarfed: minor typo: disabled (near the end)

readme.txt
```diff
@@ -86,7 +86,8 @@ Supports the full OAuth2/IndieAuth authentication and authorization flow. Defaul
 
 If the token's `me` value matches a WordPress user's URL, that user will be used. Otherwise, the token must match the site's URL, and no user will be used.
 
-Alternatively, you can set `MICROPUB_LOCAL_AUTH` to 1 to disable the plugin's authorization function, for example if you want authorization to be done by WordPress or another plugin such as the Indieauth plugin.
+Alternatively, you can set `MICROPUB_LOCAL_AUTH` to 1 to disable the plugin's authorization function, for example if you want authorization to be done by WordPress or another plugin. It will also 
+disable if the IndieAuth plugin is installed.
```
snarfed: ditto

snarfed (changes requested):
otherwise looks good!

dshanske:
Is that better?

micropub.php
```diff
@@ -84,7 +84,7 @@ public static function init() {
 		add_filter( 'webfinger_user_data', array( $cls, 'micropub_jrd_links' ) );
 
 		// Disable adding headers if local auth is set
-		if ( ! MICROPUB_LOCAL_AUTH ) {
+		if ( MICROPUB_LOCAL_AUTH  || ! class_exists( 'IndieAuth_Plugin' ) ) {
```
snarfed: huh, just noticed, you need to keep the `!` before `MICROPUB_LOCAL_AUTH`, right?

good reminder that we always need manual testing for changes that aren't unit tested. :P

snarfed: oh and should the `||` be `&&`?

dshanske: No, because it disables if the declaration is not 0. So the ! was wrong. And it should be || because if the Indieauth plugin is installed, then it should be disabled.

snarfed: hrm, that's disappointing. looks like this was added in #103, which we supposedly manually tested and confirmed that it worked. so maybe we need to manually test more or better, since it wasn't enough before? or (maybe better :P) we need to add unit tests?

dshanske: I did test it. But point taken.

snarfed: ok, i understand the logic now. thanks!

i'm a little more nervous now about being sure all this changed auth logic is right, enough that we'll probably want to do a final round of manual testing before the net release. and i'll probably want us to add unit tests the next time we make any nontrivial changes. but i'll happily merge now!

snarfed (approved):

micropub.php
dshanske: I understand. I intend to see about finishing the Indieauth work and having that out to test against. Some of the improvements and tests I may port up to Micropub.

Issue #112, opened 2018-03-26
Micropub Errors 


stephenpieper:
I've started getting Micropub errors while using different services. I noticed it with OwnYourSwarm and Quill. The errors are the same in both cases. 

<b>Warning</b>:  array_diff(): Argument #1 is not an array in <b>/...wp-content/plugins/micropub/micropub.php</b> on line <b>252</b><br />

<b>Warning</b>:  Cannot modify header information - headers already sent by (output started at /...wp-content/plugins/micropub/micropub.php:252) in <b>.../wp-content/plugins/micropub/micropub.php</b> on line <b>1092</b><br />
{"error":"invalid_request","error_description":"Invalid Post Status"}

I haven't changed anything as far as I know.

I'm using Wordpress-Micropub v1.3 with other IndieWeb plugins like Simple Location.

snarfed:
thanks for filing, and sorry for the trouble!

@dshanske, based on `Invalid Post Status`, mind helping debug?

dshanske:
I have a solution, but I can't explain why the current code isn't working.

snarfed:
looks like this is also #104.

snarfed:
@stephenpieper we think this might be fixed now. if you're comfortable installing the current plugin code in github, feel free to try it and let us know if it works for you. thanks!

stephenpieper:
I've installed the new code. The first error has gone but  but I'm still getting

{"error":"invalid_request","error_description":"Invalid Post Status"}

.

dshanske:
I am still stumped. I am going to try completely redoing the code that generates the post status. I cannot reproduce this.

stephenpieper:
I've installed the previous version of the plugin and it works so it's obviously something to do with 1.3.

snarfed:
thank you for sticking with this, @stephenpieper, and for the debugging! that's valuable info.

i've merged @dshanske's latest fix attempt into master,  #115. let's see if that fixes it. if not, we can start narrowing down based on the diff between 1.2 and 1.3.

stephenpieper:
No luck yet. I'm also getting this error with Quill if it's any use.

"Your endpoint did not return a Location header."

dshanske:
I am going to keep working on this, but still stumped and working in the dark. All my manual tests and unit tests pass.

dshanske:
We have two people reporting the problem, I wonder what the two of you have in common.

dshanske:
What can you tell me about your installation?

[...]

Issue #104, opened 2018-01-18
Invalid Post Status


andrewgribben:
`POST http://grib.co/?micropub=endpoint HTTP/1.1
Authorization: Bearer eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJtZSI6Imh0dHA6XC9cL2dyaWIuY29cLyIsImlzc3VlZF9ieSI6Imh0dHBzOlwvXC90b2tlbnMuaW5kaWVhdXRoLmNvbVwvdG9rZW4iLCJjbGllbnRfaWQiOiJodHRwczpcL1wvbWljcm9wdWIucm9ja3NcLyIsImlzc3VlZF9hdCI6MTUxNjIzNTc2NSwic2NvcGUiOiJjcmVhdGUgdXBkYXRlIGRlbGV0ZSB1bmRlbGV0ZSIsIm5vbmNlIjoxMDE5NTYyMDAxfQ.qOcvnaw4FvhhCyiVNfPvoVeMyDSiL8V78-1x5wvQzVg
Content-type: application/x-www-form-urlencoded; charset=utf-8
h=entry&
content=Micropub+test+of+creating+a+basic+h-entry`

And the response

`HTTP/1.1 400 Bad Request
Server: nginx/1.10.3 (Ubuntu)
Date: Thu, 18 Jan 2018 00:53:34 GMT
Content-Type: application/json; charset=UTF-8
Transfer-Encoding: chunked
Connection: keep-alive
Link: <https://indieauth.com/auth>; rel="authorization_endpoint"
Link: <https://tokens.indieauth.com/token>; rel="token_endpoint"
Link: <http://grib.co/?micropub=endpoint>; rel="micropub"
Link: <http://grib.co/wp-json/webmention/1.0/endpoint>; rel="webmention"
Link: <http://grib.co/wp-json/webmention/1.0/endpoint>; rel="http://webmention.org/"
{"error":"invalid_request","error_description":"Invalid Post Status"}`

I’m using nginx and PHPfpm and have added the following to the fastcgi.conf thinking that the issue was header related, but nothing seems to work.
fastcgi_param  Authorization $http_authorization;
fastcgi_param REMOTE_USER $remote_user;

I can’t create microformat posts from Quill etc but somehow managed to post from teacup this morning, but it no longer works. PHP logs show:

[...]

PR #113, opened 2018-03-28
Set default as global default is not working

dshanske:
The global default, set during register_setting, is not working for everyone. So, the pre WordPress 4.7 world, the default would be in the get_option command itself. This reverts to that behavior for now.


snarfed:
this is for #112 and #104, right?

snarfed:
looks great, thanks!

PR #115, opened 2018-04-01
Rearrange Post Status option to return default before specifics

dshanske:
I want to try returning the default post status before going into the overrides. So, if nothing is set, it will do this early on, and see if that addresses the issue.


snarfed:
man, so disappointing that we need to attempt blind fixes like this because we can't reproduce it ourselves...but sure, why not. :P

dshanske:
I am annoyed by that myself. Even the unit tests aren't tripping it.

PR #118, opened 2018-04-05
Add Support for Legacy and New Scopes

dshanske:
Right now the plugin completely ignores scopes. This gives it support for current and future scopes.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -7,7 +7,7 @@
  * Author: Ryan Barrett
  * Author URI: https://snarfed.org/
  * Text Domain: micropub
- * Version: 1.4.1
+ * Version: 1.4.2
  */
 
 /* See README for supported filters and actions.
@@ -63,6 +63,10 @@ class Micropub_Plugin {
 
 	// associative array, populated by authorize().
 	protected static $micropub_auth_response;
+
+
+	// Array of Scopes
+	protected static $scopes;
 
 	/**
 	 * Initialize the plugin.
@@ -166,6 +170,11 @@ class Micropub_Plugin {
 		// Be able to bypass Micropub auth with other auth
 		if ( MICROPUB_LOCAL_AUTH || class_exists( 'IndieAuth_Plugin' ) ) {
 			$user_id = get_current_user_id();
+			// The WordPress IndieAuth plugin uses a global variable named $indieauth_scopes to store this info
+			global $indieauth_scopes;
+			if ( ! empty( $indieauth_scopes ) ) {
+				static::$scopes = $indieauth_scopes;
+			}
 			if ( ! $user_id ) {
 				static::handle_authorize_error( 401, 'Unauthorized' );
 			}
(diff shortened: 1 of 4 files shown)


snarfed:
this is really great, i'm glad we're adding scope support. out of curiosity though, why add a filter? what's the use case for other plugins or themes overriding scope auth checks?

dshanske:
New scopes.

dshanske:
@snarfed Rebased this based on the other merges. I'm willing to remove the filter, but I think it would help optional expansion. Will defer.

snarfed:
yeah, let's drop the filter for now. i'd need to hear at least a couple different concrete use cases for extensible scope authorization before it would make sense to me personally, and maybe not even then. auth is important and delicate, and not generally something you want to let any other code, written by anyone, arbitrarily override.

aaronpk:
Agreed on removing the filter. Until there is a good use case for a plugin modifying the behavior of the plugin that also wants to provide finer-grained access, there's no need to make that part extensible. Supporting the primary list of scopes create/update/delete/undelete is good for now, and you could also accept "post" if it's requested for legacy clients.

dshanske:
@snarfed You suggested that I removed post scope. I didn't, I just moved the check further down into the capability function. It treats post as full permission, and create, update, delete, or undelete as the scopes for the individual functions... although with only a few clients supporting undelete, maybe delete should be the scope for either.

snarfed:
true! you're absolutely right, sorry about that. thanks!

snarfed (approved):
lgtm! just one refactoring suggestion. feel free to merge after that!

dshanske:
I am thinking that I might make some other changes now that I am not using a filter. Move the WordPress permission check into the check scope function. May ask for your thoughts once done.

micropub.php
```diff
@@ -246,6 +250,19 @@ private static function authorize() {
...
 
+	/**
+	 * Check scope
+	 *
+	 * @param string $scope
+	 *
+	 * @return boolean
+	**/
+	protected static function check_scope( $scope ) {
+		if ( in_array( $scope, static::$scope, true ) || in_array( 'post', static::$scope, true ) ) {
```
snarfed: very minor nit: this can now be just `return in_array(...) || in_array(...);`

readme.md
```diff
@@ -17,6 +17,18 @@ Supports the [full W3C Micropub CR spec](https://www.w3.org/TR/micropub/) as of
 This project is placed in the public domain. You may also use it under the [CC0 license](http://creativecommons.org/publicdomain/zero/1.0/).
 
 
+## Scope 
+
+Supports the following [scope](https://indieweb.org/scope) parameters requested by Micropub clients.
+* Post (legacy) - Grants all user delegated access
```
snarfed: these should probably be lower case, eg `post`, since they're case sensitive.

readme.txt
```diff
@@ -25,6 +25,17 @@ Supports the [full W3C Micropub CR spec](https://www.w3.org/TR/micropub/) as of
 
 This project is placed in the public domain. You may also use it under the [CC0 license](http://creativecommons.org/publicdomain/zero/1.0/).
 
+== Scope ==
+
+Supports the following [scope](https://indieweb.org/scope) parameters requested by Micropub clients.
```
snarfed: ditto

snarfed (commented):
looks good! only a couple minor nits.

btw, github evidently doesn't notify a PR's reviewers etc when you just amend an existing commit like this. it also makes reviewing a bit harder, since it's hard to tell just what's changed since you last reviewed. might be one reason to wait on squashing until after reviewing. just fyi.

PR #119, opened 2018-04-06
Improve Spec Compliance

dshanske:
https://www.w3.org/TR/micropub/#syndication-targets-p-4

Syndication targets must have a UID and a Name. 

https://www.w3.org/TR/micropub/#examples-of-creating-objects-p-4

Property should be mp-syndicate-to not syndicate-to. 

https://www.w3.org/TR/micropub/#syndication-targets

The query only is syndicate-to 

https://indieweb.org/Micropub-extensions

The extension for slug is mp-slug at the moment. Updating to comply with same.


snarfed:
looks great! thank you!

PR #120, opened 2018-04-06
Add support for querying categories

dshanske:
This returns a merged list of tags and categories if q=category is sent to the endpoint.


snarfed:
looks good! mind adding a test?

snarfed:
thanks!

PR #121, opened 2018-04-07
Add syndication hook

dshanske:
This PR adds a hook that only fires if there are verified syndication requests. This makes it easier for a third-party plugin to hook in, rather than the more generic after_micropub which covers everything.

The goal is to be able to say to a POSSE plugin developer, add your service to this filter, and trigger POSSE on this action.


snarfed:
great idea! mind adding a test?

dshanske:
@snarfed Test added.

snarfed:
thanks!

PR #122, opened 2018-04-07
Add support for location-visibility altitude and accuracy

dshanske:
Location-visibility is a proposed property that maps to geo_public, a known WordPress Geodata property. I currently use this in my Simple Location plugin.

Altitude was already in your h-geo test data, and updated to store this property. For Geo URIs, added altitude if present, and storing uncertainty as accuracy.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -885,13 +885,43 @@ class Micropub_Plugin {
 	 * https://ownyourswarm.p3k.io/docs#checkins
 	 *
 	 * WordPress geo data is stored in post meta: geo_address (free text),
-	 * geo_latitude, and geo_longitude:
+	 * geo_latitude, geo_longitude, and geo_public:
 	 * https://codex.wordpress.org/Geodata
+	 * It is noted that should the HTML5 style geolocation properties of altitude, accuracy, speed, and heading are
+	 * used they would use the same geo prefix. Simple Location stores these when available using accuracy to estimate
+	 * map zoom when displayed.
 	 */
 	public static function store_geodata( $args ) {
 		$properties = static::get( static::$input, 'properties' );
 		$location   = static::get( $properties, 'location', static::get( $properties, 'checkin' ) );
 		$location   = static::get( $location, 0, null );
+		// Location-visibility is an experimental property https://indieweb.org/Micropub-extensions#Location_Visibility
+		// It attempts to mimic the geo_public property
+		$visibility = static::get( $properties, 'location-visibility', null );
+		if ( $visibility ) {
+			$visibility = array_pop( $visibility );
+			if ( ! isset( $args['meta_input'] ) ) {
+				$args['meta_input'] = array();
+			}
+			switch ( $visibility ) {
+				// Currently supported by https://github.com/dshanske/simple-location as part of the Geodata store noted in codex link above
+				// Public indicates coordinates, map, and textual description displayed
+				case 'public':
+					$args['meta_input']['geo_public'] = 1;
+					break;
+				// Private indicates no display
+				case 'private':
+					$args['meta_input']['geo_public'] = 0;
+					break;
+				// Protected which is not in the original geodata spec is used by Simple Location to indicate textual description only
+				case 'protected':
+					$args['meta_input']['geo_public'] = 2;
+					break;
+				default:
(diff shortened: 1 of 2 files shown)


micropub.php
```diff
@@ -882,13 +882,36 @@ public static function default_file_handler( $post_id ) {
...
+	 * It is noted that should the HTML5 style geolocation properties of altitude, accuracy, speed, and heading are
+	 * used they would use the same geo prefix. Simple Location stores these when available using accuracy to estimate
+	 * map zoom when displayed.
 	 */
 	public static function store_geodata( $args ) {
 		$properties = static::get( static::$input, 'properties' );
 		$location   = static::get( $properties, 'location', static::get( $properties, 'checkin' ) );
 		$location   = static::get( $location, 0, null );
+		// Location-visibility is an experimental property https://indieweb.org/Micropub-extensions#Location_Visibility
+		// It attempts to mimic the geo_public
```
snarfed: incomplete sentence? ok if this was intentional, just checking.

micropub.php
```diff
@@ -882,13 +882,36 @@ public static function default_file_handler( $post_id ) {
...
+				case 'public':
+					$args['meta_input']['geo_public'] = 1;
+					break;
+				case 'private':
+					$args['meta_input']['geo_public'] = 0;
+					break;
+				case 'protected':
+					$args['meta_input']['geo_public'] = 2;
+					break;
+			}
```
snarfed: we should probably return an error if it's something other than these three values, right?

micropub.php
```diff
@@ -882,13 +882,36 @@ public static function default_file_handler( $post_id ) {
...
+		// It attempts to mimic the geo_public
+		$visibility = static::get( $properties, 'location-visibility', null );
+		if ( $visibility ) {
+			$visibility = array_pop( $visibility );
+			if ( ! isset( $args['meta_input'] ) ) {
+				$args['meta_input'] = array();
+			}
+			switch ( $visibility ) {
+				case 'public':
+					$args['meta_input']['geo_public'] = 1;
```
snarfed: mind adding a comment with a link to wordpress docs about what these numeric values mean? or even better, are there constants in the wordpress code that we could use instead?

tests/test_micropub.php
```diff
@@ -490,17 +491,46 @@ function test_create_location_h_adr() {
 		$this->assertEquals( '', get_post_meta( $post->ID, 'geo_address', true ) );
 	}
 
-	function test_create_location_plain_text() {
```
snarfed: did you mean to delete the `_plain_text` test? we still want it, right?

snarfed (changes requested):
thank you for the detailed comments and test! this is great. nothing major, just a few minor things.

micropub.php
```diff
@@ -917,17 +940,24 @@ public static function store_geodata( $args ) {
 				}
 				$args['meta_input']['geo_latitude']  = $props['latitude'][0];
 				$args['meta_input']['geo_longitude'] = $props['longitude'][0];
+				$args['meta_input']['geo_altitude']  = $props['altitude'][0];
 			} elseif ( 'geo:' === substr( $location, 0, 4 ) ) {
 				// Geo URI format:
 				// http://en.wikipedia.org/wiki/Geo_URI#Example
-				// https://indiewebcamp.com/micropub##location
+				// https://indieweb.org/micropub##location
```
snarfed: the wiki doesn't support fragmentions, so this link doesn't work. let's use https://indieweb.org/Micropub#h-entry instead.

micropub.php
dshanske: It is in the codex link referenced. However, will comment about what they mean inline.

tests/test_micropub.php
dshanske: Accident. Restoring it.

micropub.php
snarfed: i meant the `location-visibility` micropub property values. https://indieweb.org/Micropub-extensions#Location_Visibility only lists `public`, `protected`, and `private`, so we should probably return 400 if we get anything else.

dshanske:
@snarfed Redid this one.

dshanske:
@snarfed Will reject any unapproved property now.

snarfed:
yay!

PR #127, opened 2018-04-26
Pull token data from IndieAuth plugin

dshanske:
I noticed that when I use the IndieAuth endpoint, the auth response storage vanished. This restores that behavior.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -64,7 +64,6 @@ class Micropub_Plugin {
 	// associative array, populated by authorize().
 	protected static $micropub_auth_response;
 
-
 	// Array of Scopes
 	protected static $scopes;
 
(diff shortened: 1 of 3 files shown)


micropub.php
```diff
@@ -175,6 +174,10 @@ public static function parse_query( $wp ) {
 			if ( ! empty( $indieauth_scopes ) ) {
 				static::$scopes = $indieauth_scopes;
 			}
+			global $indieauth_token;
```
snarfed: huh. is this really an ok way for plugins to communicate?! seems like they should use hooks or filters instead?

snarfed: i do get that the code was already working this way, and this just extends that. still though.

dshanske: I am happy to see explore changibg it to a filter on both ends

snarfed: thanks! if you plan to switch to filters, feel free to merge this now and do that afterward.

dshanske:
@snarfed Switches to filters.

PR #134, opened 2018-07-09
Split endpoint into multiple files and add dedicated settings page

dshanske:
This splits  the micropub.php file into three files

- Admin, which contains a new settings page where previously the sole setting was added to the writing settings page and often overlooked. In a future PR I want to add the IndieAuth settings that are no longer present in the Indieauth plugin, the ability to change the endpoint from indieauth.com
- Render, which contains the rendering code, which no longer has any direct connection to the rest of the code. This will allow future enhancements as previously proposed.
- Endpoint, which will contain the endpoint code.

This is part of a series of PRs I've been working on. But this change makes it easier for me to start building the new structure.


includes/class-micropub-admin.php
```diff
@@ -0,0 +1,93 @@
...
+			array( $cls, 'default_post_status_setting' ),
+			$page,
+			'micropub_writing'
+		);
+	}
+
+	/**
+	 * Add admin menu entry
+	 */
+	public static function admin_menu() {
```
snarfed: this is new, right? in the future, especially for big structural changes like this (or refactorings etc), you generally want to _just_ make the structural change, in its own commit or PR, _without_ functionality changes, and then make functionality changes in a separate commit or PR. makes reviewing easier.

snarfed:
thanks for doing this! really great. much appreciated.

hard to review, since there are no diffs, so note to myself: i ended up copying the old `micropub.php` to `/tmp` and running `diff ~/micropub.php includes/class-micropub-endpoint.php`, etc. to help see the changes.

includes/class-micropub-endpoint.php
```diff
@@ -0,0 +1,1201 @@
...
+		$option = get_option( 'micropub_default_post_status', '' );
+		return empty( $option ) ? ( MICROPUB_DRAFT_MODE ? 'draft' : 'publish' ) : $option;
+	}
+
+	private static function post_status( $mf2 ) {
+		$props = $mf2['properties'];
+		// If both are not set immediately return
+		if ( ! isset( $props['post-status'] ) && ! isset( $props['visibility'] ) ) {
+			$status = self::default_post_status();
+			if ( ! in_array( $status, array( 'publish', 'draft', 'private' ), true ) ) {
```
snarfed: looks like this changed. old code is below. change looks fine, except this `if` (lines 538-540) is entirely a noop, since both paths `return $status`. mind deleting it?

```php
private static function default_post_status() {
	return MICROPUB_DRAFT_MODE ? 'draft' : 'publish';
}

private static function post_status( $mf2 ) {
	$props = $mf2['properties'];
	// If both are not set immediately return
	if ( ! isset( $props['post-status'] ) && ! isset( $props['visibility'] ) ) {
		$status = get_option( 'micropub_default_post_status', static::default_post_status() );
		if ( ! in_array( $status, array( 'publish', 'draft', 'private' ), true ) ) {
			return static::default_post_status();
		}
		return $status;
	}
	...
```

includes/class-micropub-render.php
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Micropub Render Class
+ */
+class Micropub_Render {
+	/**
+	 * Generates and returns a post_content string suitable for wp_insert_post()
+	 * and friends.
+	 */
+	public static function generate_post_content( $post_content, $input ) {
```
snarfed: this function and `generate_event` are still in `class-micropub-endpoint.php`. they shouldn't be, right? mind removing them from there?

templates/micropub-writing-settings.php
```diff
@@ -1,10 +0,0 @@
-<fieldset id="micropub">
```
snarfed: in the future, if you move a file, you want to use `git mv` so that git records the move, so you can see diffs, follow history across filenames, etc.

dshanske: I git mved the wrong file.

includes/class-micropub-endpoint.php
dshanske: I made some adjustments to this to move more of the logic into the default function.

dshanske:
@snarfed This hopefully addresses the requested changes. Already planning the next round.

snarfed (approved):
looks great! thanks again!

PR #135, opened 2018-07-11
Split off Authorization

dshanske:
This PR splits authorization into its own file. It also changes the load process for it by making it use the WordPress login functionality of get_current_user_id, which calls wp_get_current_user which calls the determine_current_user filter

As one change, if it cannot determine a user ID, it will not let the post  be made, as it cannot match it up with an existing user. I think this can be addressed by improving user_url, which I intend to update in a future PR to include the improvements made in the similar IndieAuth function...matching the default user if set by the  IndieWeb plugin, checking to see if there is only a single user, or only a single user with author permissions, etc. Also, allowing the author URL(/author/snarfed) as an initial URL.) 

As part of  the changes I'm trying to make, I'm  trying to move the  code to work more within the WordPress core code, and trying to divide this work into pieces.

### DIFF
--- a/includes/class-micropub-authorize.php
+++ b/includes/class-micropub-authorize.php
@@ -165,7 +165,9 @@ class Micropub_Authorize {
 		// look for a user with the same url as the token's `me` value.
 		$user = static::user_url( $me );
 
-		// IndieAuth Plugin uses priority 9		
+		// IndieAuth Plugin uses priority 9
+		// TODO: These filters are added here to ensure that they are loaded after the scope is set
+		// However future changes will change the order so this is no longer needed
 		add_filter( 'indieauth_scopes', array( $cls, 'indieauth_scopes' ), 11 );
 		add_filter( 'indieauth_response', array( $cls, 'indieauth_response' ), 11 );
 
--- a/includes/class-micropub-authorize.php
+++ b/includes/class-micropub-authorize.php
@@ -19,11 +19,11 @@ add_action( 'plugins_loaded', array( 'Micropub_Authorize', 'init' ) );
  */
 class Micropub_Authorize {
 
-	// associative array, populated by authorize().
-	protected static $micropub_auth_response;
+	// associative array, populated by determine_current_user.
+	protected static $micropub_auth_response = array();
 
 	// Array of Scopes
-	protected static $scopes;
+	protected static $scopes = array();
 
 	/**
 	 * Initialize the plugin.
@@ -37,6 +37,10 @@ class Micropub_Authorize {
 		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
 		// The WordPress IndieAuth plugin uses priority 30
 		add_filter( 'determine_current_user', array( $cls, 'determine_current_user' ), 31 );
+
+		// IndieAuth Plugin uses priority 9		
+		add_filter( 'indieauth_scopes', array( $cls, 'indieauth_scopes' ), 11 );
+		add_filter( 'indieauth_response', array( $cls, 'indieauth_response' ), 11 );
 	}
 
 	public static function indieauth_scopes( $scopes ) {
@@ -119,7 +123,7 @@ class Micropub_Authorize {
 			return $user_id;
 		}
 
-		// Only try to authenticate if this is the Micropub endpoint
+		// Since this runs on the built-in determine_current user filter only try to authenticate if this is the Micropub endpoint
 		if ( ! empty( get_query_var( 'micropub' ) ) ) {
 			return $user_id;
 		}
@@ -164,8 +168,6 @@ class Micropub_Authorize {
 		// look for a user with the same url as the token's `me` value.
 		$user = static::user_url( $me );
 
-		add_filter( 'indieauth_scopes', array( $cls, 'indieauth_scopes' ), 11 );
-		add_filter( 'indieauth_response', array( $cls, 'indieauth_response' ), 11 );
 		if ( $user ) {
 			return $user;
 		}
--- a/includes/class-micropub-authorize.php
+++ b/includes/class-micropub-authorize.php
@@ -1,0 +1,218 @@
+<?php
+
+/* For debugging purposes this will bypass Micropub authentication
+ * in favor of WordPress authentication
+ * Using this to test querying(q=) parameters quickly
+ */
+// Allows for a custom Authentication and Token Endpoint
+if ( ! defined( 'MICROPUB_AUTHENTICATION_ENDPOINT' ) ) {
+	define( 'MICROPUB_AUTHENTICATION_ENDPOINT', 'https://indieauth.com/auth' );
+}
+if ( ! defined( 'MICROPUB_TOKEN_ENDPOINT' ) ) {
+	define( 'MICROPUB_TOKEN_ENDPOINT', 'https://tokens.indieauth.com/token' );
+}
+
+add_action( 'plugins_loaded', array( 'Micropub_Authorize', 'init' ) );
+
+/**
+ * Micropub IndieAuth Authorization Class
+ */
+class Micropub_Authorize {
+
+	// associative array, populated by authorize().
+	protected static $micropub_auth_response;
+
+	// Array of Scopes
+	protected static $scopes;
+
+	/**
+	 * Initialize the plugin.
+	 */
+	public static function init() {
+		$cls = get_called_class();
+
+		add_action( 'wp_head', array( $cls, 'html_header' ), 99 );
+		add_action( 'send_headers', array( $cls, 'http_header' ) );
+		add_filter( 'host_meta', array( $cls, 'jrd_links' ) );
+		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
+		// The WordPress IndieAuth plugin uses priority 30
+		add_filter( 'determine_current_user', array( $cls, 'determine_current_user' ), 31 );
+	}
+
+	public static function indieauth_scopes( $scopes ) {
+		return static::$scopes;
+	}
+	public static function indieauth_response( $response ) {
+		return static::$micropub_auth_response;
+	}
+
+	public static function header( $header, $value ) {
+			header( $header . ': ' . $value, false );
+	}
+
+	public static function http_header() {
+			static::header( 'Link', '<' . get_option( 'indieauth_authorization_endpoint', MICROPUB_AUTHENTICATION_ENDPOINT ) . '>; rel="authorization_endpoint"' );
+			static::header( 'Link', '<' . get_option( 'indieauth_token_endpoint', MICROPUB_TOKEN_ENDPOINT ) . '>; rel="token_endpoint"' );
+	}
+	public static function html_header() {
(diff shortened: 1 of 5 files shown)


snarfed:
woo, thanks!

note to myself: reviewing the new file with `diff /tmp/class-micropub-endpoint.php /tmp/class-micropub-authorize.php`, where `class-micropub-endpoint.php` is from master.

includes/class-micropub-authorize.php
```diff
@@ -0,0 +1,218 @@
...
+	 * Initialize the plugin.
+	 */
+	public static function init() {
+		$cls = get_called_class();
+
+		add_action( 'wp_head', array( $cls, 'html_header' ), 99 );
+		add_action( 'send_headers', array( $cls, 'http_header' ) );
+		add_filter( 'host_meta', array( $cls, 'jrd_links' ) );
+		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
+		// The WordPress IndieAuth plugin uses priority 30
```
snarfed: these kinds of comments are _great_. thank you!

micropub.php
```diff
@@ -28,11 +28,20 @@
  * 5. Extract the access_token parameter from the response body.
  */
 
+if ( ! defined( 'MICROPUB_LOCAL_AUTH' ) ) {
+	        define( 'MICROPUB_LOCAL_AUTH', '0' );
```
snarfed: indentation

snarfed:
also, again, this has significant changes in functionality, which ideally should happen in a separate PR from splitting files.

...and, now that we're making even more changes to how auth works, we should really start adding auth unit tests. :P not strictly necessary for this PR, but especially since you have a lot more changes to send, i'd like to see auth tests sooner rather than later. (thank you in advance!)

includes/class-micropub-authorize.php
```diff
@@ -0,0 +1,218 @@
...
+
+	public static function get( $array, $key, $default = array() ) {
+		if ( is_array( $array ) ) {
+			return isset( $array[ $key ] ) ? $array[ $key ] : $default;
+		}
+		return $default;
+	}
+
+	/**
+	 * Get the authorization header
```
snarfed: this is so great! thank you for adding it! it's probably been the single biggest pain point for users for a while now. did you get the code from @zegnat?

it's probably worth a unit test. mind adding one? and also updating the docs, specifically the _Frequently Asked Questions_ section in the readme?

includes/class-micropub-authorize.php
```diff
@@ -0,0 +1,218 @@
...
+			);
+		}
+
+		$me                             = untrailingslashit( $params['me'] );
+		static::$micropub_auth_response = $params;
+
+		// look for a user with the same url as the token's `me` value.
+		$user = static::user_url( $me );
+
+		add_filter( 'indieauth_scopes', array( $cls, 'indieauth_scopes' ), 11 );
```
snarfed: it seems a bit surprising that these happen here, lazily during a request, instead of in `init`. mind moving them there?

includes/class-micropub-authorize.php
```diff
@@ -0,0 +1,218 @@
...
+		add_filter( 'indieauth_response', array( $cls, 'indieauth_response' ), 11 );
+		if ( $user ) {
+			return $user;
+		}
+
+		// no user with that url. if the token is for this site itself, allow it and
```
snarfed: in the detailed PR comment (thank you for that btw!), you said:

> As one change, if it cannot determine a user ID, it will not let the post be made, as it cannot match it up with an existing user.

...but this seems to still allow posts without a user...?

includes/class-micropub-authorize.php
```diff
@@ -0,0 +1,218 @@
...
+	 */
+	public static function determine_current_user( $user_id ) {
+		$cls = get_called_class();
+		// Do not try to find a user if one has already been found
+		if ( ! empty( $user_id ) ) {
+			return $user_id;
+		}
+
+		// Only try to authenticate if this is the Micropub endpoint
+		if ( ! empty( get_query_var( 'micropub' ) ) ) {
```
snarfed: very minor nit: for readability, consider adding something like:

> ..., since this runs as a filter on the built in `determine_current_user` function.

as is, it took me a few minutes to figure out why we had this. 😁

snarfed (commented):
a few small things, nothing big. this is great!

includes/class-micropub-authorize.php
dshanske: This is something I put together after reading dozens of AUTHORIZATION header error messages in a Google Search. It tries the $_SERVER first before looking for getallheaders, which is an Apache only function. It isn't perfect though. I use a similar function in the IndieAuth plugin.

includes/class-micropub-authorize.php
dshanske: get_current_user_id should return 0. This returns null. The reason for this difference is to allow better error handling. It also optionally returns an error function.

But, if you look at where it returns, if ! $user_id then it returns unauthorized

includes/class-micropub-authorize.php
dshanske: As for unit tests, I'd like to build unit tests for this and the IndieAuth plugin, as they are both acting as clients and can both use the same tests, or variants thereof. I actually want to bring some of the other improvements over.

dshanske: I have another PR planned to enhance the user_url functions with what I learned in the other plugin and will be submitting it with the additional unit tests, if you feel comfortable merging as I have the start of the unit tests in that code.

dshanske:
I think I remembered why I put that add_action inline...as before that, it would not have been set.

dshanske:
@snarfed Can you have a look?

includes/class-micropub-authorize.php
snarfed: > I'd like to build unit tests for this and the IndieAuth plugin

i like that instinct toward reuse! sharing a set of unit tests across multiple different codebases is very difficult at best, and probaby impossible in most cases, like this. (i'm not sure i've _ever_ seen it done.) we should plan to just write unit tests for this plugin here, like normal.

(compliance test suites like https://micropub.rocks/ are great, of course, and obviously can be used with mutliple codebases! they're applications themselves though, not unit tests. unit tests are different.)

snarfed: also we should definitely still ask @zegnat to look at this (without blocking the PR) and maybe contribute his work too! he did a great deep dive on this a bit ago.

..aha, here: https://gist.github.com/Zegnat/f4192c3cce05d847af3547eea80e383c

includes/class-micropub-authorize.php
snarfed: ping!

micropub.php
snarfed: ping

includes/class-micropub-authorize.php
snarfed: huh. so if i understand you right, this block is depending on the fact that `$user_id` here is 0 and `$user` is null? if so, ok, but then change the `return`s below to `return 0` and `return null`, or whatever, explicitly. _much_ more readable.

includes/class-micropub-authorize.php
dshanske: Both plugins need different unit tests, but the ones for testing auth headers would be the same. Either way, I'd like to get @Zegnat 's opinion, and maybe incorporate a version of the test code that would allow checking headers in one-click.

dshanske:
I removed the duplicate return which should clarify this.

includes/class-micropub-authorize.php
Zegnat: Looks like my name has been said three times, so it is time for me to show up. This seems to be as complete as you are going to get it, and very close to what I am [doing in Mintoken](https://github.com/Zegnat/php-mintoken/blob/f93b888c4c56dd5faa129b6c0ccae90b7b00132f/endpoint.php#L196-L202). In fact, I think I am going to adapt the `REDIRECT_HTTP_AUTHORIZATION` into Mintoken as well! That’s a great idea!

The only real choices you get to make in this implementation are:

1. In what order do you check? (`$_SERVER` vs. server-specific functions.) I think checking `$_SERVER` first is valid and saves you a function check on all non-apache servers, though such micro-optimisations are probably worthless in the grand scheme of a WordPress extension.
   
2. Do you use `getallheaders`, `apache_request_headers`, or both, for the server-specific function? For the diagnostics tool I check both, but so far have not run into any servers that supported one without the other. That’s why I only use one in Mintoken.
   
   I choose to use `apache_request_headers` because it clearly illustrates that it isn’t going to work anywhere else. But that is a personal preference thing more than anything else.

:+1: from me.

micropub.php
pfefferle: :)

includes/class-micropub-authorize.php
dshanske: getallheaders is an Apache  function, if I remember correctly, is an alias for apache_request_headers.(https://secure.php.net/manual/en/function.getallheaders.php). That is why, if it isn't defined, the built-in version takes over, which effectively does the same thing. The reasoning for checking $_SERVER first is more that, is that PHP itself can continue to improve how it loads the server environment variable. It isn't about the optimization so much as allowing other code that may get upgraded in future to try.

pfefferle: Have you seen/tried: https://developer.wordpress.org/reference/classes/wp_rest_request/get_header/

Zegnat: > Have you seen/tried: https://developer.wordpress.org/reference/classes/wp_rest_request/get_header/

Looking at that it relies 100% on an array that it has previously been fed through `set_headers`. This happens [in the WP_REST_Server class](https://github.com/WordPress/WordPress/blob/4c2fcf36a3193a7e63a82e5ff459f2587c69a859/wp-includes/rest-api/class-wp-rest-server.php#L309), but relies only on `$_SERVER` (no Apache fallback) and uses [a standard looping method](https://github.com/WordPress/WordPress/blob/4c2fcf36a3193a7e63a82e5ff459f2587c69a859/wp-includes/rest-api/class-wp-rest-server.php#L1315-L1342) to only extract keys starting with `HTTP_` (missing the `REDIRECT_HTTP_AUTHORIZATION` key).

includes/class-micropub-authorize.php
```diff
@@ -0,0 +1,215 @@
...
+
+		// look for a user with the same url as the token's `me` value.
+		$user = static::user_url( $me );
+
+		// IndieAuth Plugin uses priority 9		
+		add_filter( 'indieauth_scopes', array( $cls, 'indieauth_scopes' ), 11 );
+		add_filter( 'indieauth_response', array( $cls, 'indieauth_response' ), 11 );
+
+		if ( $user ) {
+			return $user;
```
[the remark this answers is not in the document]
dshanske: user_url returns the user ID  from the URL, so $user and $user_id should both be IDs.

dshanske:
I am making a new commit to break out the token code so I can unit test more of it.

dshanske:
So, reverted the last change and added the notation re future actions. This is going to take a few more PRs till version 2.0 is ready.

snarfed (approved):

PR #136, opened 2018-07-14
Create Media Endpoint

dshanske:
This adds a media endpoint with unit tests. It uses the WordPress REST API and built-in WordPress functionality. I plan to rewrite the micropub endpoint to use the same REST API building blocks.


dshanske:
@snarfed This requires the new authorization piece or the Indieauth plugin to work, FYI

dshanske:
But.... unit tests!!

dshanske:
Rebased for authorization branch

includes/class-micropub-media.php
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	/**
+	* Check scope
+	*
+	* @param array $scope
+	*
+	* @return boolean
+	**/
+	protected static function check_scopes( $scopes ) {
```
snarfed: minor nit: based on this method name, it surprised me that this only checks for intersection and not superset. consider either renaming to `has_any_scope`, or probably better, just inline it on line 85 since it's only called there.

includes/class-micropub-media.php
```diff
@@ -0,0 +1,150 @@
...
+				return new WP_Micropub_Error( 'invalid_request', 'Database Error On Upload', 500 );
+			} else {
+				return new WP_Micropub_Error( 'invalid_request', $id->get_error_message(), 400 );
+			}
+		}
+
+		$attachment = get_post( $id );
+
+		// Include admin functions to get access to wp_generate_attachment_metadata().
+		require_once ABSPATH . 'wp-admin/includes/admin.php';
```
snarfed: minor nit: i prefer to do these globally, not inside functions or method. mind moving this to the top?

tests/test_media.php
```diff
@@ -0,0 +1,98 @@
...
+		wp_set_current_user( self::$author_id );
+		$response = rest_get_server()->dispatch( self::upload_request() );
+		$data     = $response->get_data();
+		$this->assertEquals( 201, $response->get_status(), wp_json_encode( $data ) );
+		// Test that a valid URL is returned in the JSON Body
+		$this->assertNotEquals( 0, attachment_url_to_postid( $data['url'] ), sprintf( '%1$s is not an attachment', $data['url'] ) );
+		// Test that a valid URL is returned in the Location Header
+		$headers = $response->get_headers();
+		$this->assertNotEquals( 0, attachment_url_to_postid( $headers['Location'] ), sprintf( '%1$s is not an attachment', $headers['Location'] ) );
+
```
snarfed: probably also check filename, mime type, title, etc?

tests/test_media.php
```diff
@@ -0,0 +1,98 @@
+<?php
+/* Media Endpoint and Upload Tests inspired by REST API Attachment Endpoint */
+
+class Micropub_Media_Test extends WP_UnitTestCase {
```
snarfed: awesome!

maybe also consider testing a request that's missing a file.

snarfed (approved):
this is really great! thank you!

the new unit test approach looks great. if you like using the REST API and request objects and testing like this, and you want to migrate the rest of the code, i'm all for it.

as mentioned, i don't actually know the REST API codebase, so i'd love @pfefferle to review this too before merge. but lgtm otherwise!

includes/class-micropub-media.php
dshanske: The problem is that globally loads it for the entire WordPress system every time. For some reason, these functions are deliberately not loaded globally at this point in time. They do the exact same thing. I based this function on one built into WordPress itself for their REST API attachment endpoint... https://developer.wordpress.org/reference/classes/wp_rest_attachments_controller/upload_from_file/

tests/test_media.php
dshanske: Will do on that.

tests/test_media.php
dshanske: Also will do.

dshanske:
Just awaiting @pfefferle to weigh in before merging.

includes/class-micropub-media.php
snarfed: interesting, ok. out of curiosity, does it cause a problem if you load it globally?

if so, and if this really is the right way, probably add a comment saying what you said here.

includes/class-micropub-endpoint.php
```diff
@@ -203,6 +203,10 @@ public static function post_handler( $user_id ) {
 		static::respond( $status, null, $args );
 	}
 
+	private static function get_syndicate_targets( $user_id ) {
```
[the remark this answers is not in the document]
snarfed: ping

includes/class-micropub-media.php
snarfed: ping

dshanske: I fixed this one

snarfed: ah, you did! i missed it since this method is still here, just now empty and unused. i expect you'll want to remove it. 😁

pfefferle (approved):
Looks good to me! The Error handling is similar to the IndieAuth plugin and helps identifying the concrete error. The REST API looks also good to me so far.

👍

I haven't tested the functionality, I only reviewed the code.

PR #138, opened 2018-07-17
Improve URL to User Function

dshanske:
Made some improvements based on the similar function in IndieAuth and added some unit tests.


snarfed:
woo tests! and new code. :P does this fix a specific bug? or just generally more functionality?

i wish there were an easy way to reuse this in place from the indie auth plugin, instead of copying and pasting. ah well.

snarfed (approved):

PR #140, opened 2018-07-18
Refactor

dshanske:
This PR removes 2 duplicate functions now in a different file and it splits out form to json conversion into its own function.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -37,13 +37,6 @@ class Micropub_Endpoint {
 		add_filter( 'host_meta', array( $cls, 'micropub_jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'micropub_jrd_links' ) );
 
-		// Disable adding headers if local auth is set
-		if ( MICROPUB_LOCAL_AUTH || ! class_exists( 'IndieAuth_Plugin' ) ) {
-			add_action( 'wp_head', array( $cls, 'indieauth_html_header' ), 99 );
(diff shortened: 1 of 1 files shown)


includes/class-micropub-endpoint.php
```diff
@@ -895,21 +891,7 @@ protected static function load_input() {
 		} elseif ( ! $content_type ||
 			   'application/x-www-form-urlencoded' === $content_type ||
 			   'multipart/form-data' === $content_type ) {
-			static::$input = array();
```
snarfed: this got lost in the move. it should be added back at the top of `form_to_json`, right?

snarfed:
hi @dshanske! i don't see how the `indieauth_*_header` methods are duplicated. i only see them defined once, which you're remove in this PR.

you don't mean they're duplicated in the indieauth plugin, do you?

dshanske:
They are in the authorize class also

includes/class-micropub-endpoint.php
dshanske: It should be $input=array() but yes

snarfed:
> They are in the authorize class also

ah, yes, thanks. i was naively looking for exact name matches, but they don't have the `indieauth_` prefix there. got it.

snarfed:
if you want to remove those methods, you need to remove their `add_action` and `add_filter`s too:

https://github.com/dshanske/wordpress-micropub/blob/800fb11206191c75fa637a5188d1fe574882d45f/includes/class-micropub-endpoint.php#L40-L46

dshanske:
@snarfed Fixed now.

snarfed (approved):
thanks!

PR #142, opened 2018-07-22
Add tests that specifically check rendering.

dshanske:
I have additional plans to break the testing apart in this manner, but wanted to submit this for feedback before I went further. 

The tests for rendering microformats test the entire process from start to finish. But as displaying microformats was broken out subsequent to their writing to be triggered by the micropub_post_content filter, it makes more sense in my opinion to just test this filter by itself.

As a positive note, this would mean, if you wanted to add things dynamically, you could hook the same function into the content filter and add microformats on the fly, or other configurations....which is why I want to test this way.

My intention is to work similarly for the geodata tests. 

This moves me closer to the previously stated goals.

### DIFF
--- a/tests/test_render.php
+++ b/tests/test_render.php
@@ -1,0 +1,182 @@
+<?php
+
+/** Unit tests for the Micropub Rendering class.
+ */
+
+class MicropubRenderTest extends WP_UnitTestCase {
+
+	public static function setUpBeforeClass() {
+		WP_UnitTestCase::setUpBeforeClass();
+	}
+
+	public function setUp() {
+		parent::setUp();
+	}
+
+	function test_create_checkin_autogenerates_checkin_text_with_content() {
+		$input = array(
+			'type' => array( 'h-entry' ),
+			'properties' => array(
+				'content' => array( 'something' ),
+				'checkin' => array( array(
+					'properties' => array(
+						'name' => array( 'Place' ),
+						'url' => array( 'http://place' ),
+					),
+				) ),
+			),
+		);
+		$post_content = Micropub_Render::generate_post_content( 'something', $input );
+
+		$this->assertEquals( "<p>Checked into <a class=\"h-card p-location\" href=\"http://place\">Place</a>.</p>\n" .
+			"<div class=\"e-content\">\nsomething\n</div>",
+			$post_content );
+	}
+
+/*
+	function check_create_content_html() {
+		$post = $this->check_create();
+		$this->assertEquals( 'HTML content test', $post->post_title );
+		// check that HTML in content isn't sanitized
+		$this->assertEquals( "<div class=\"e-content\">\n<h1>HTML content!</h1><p>coolio.</p>\n</div>", $post->post_content );
+	}
+ */
+	function _test_create_interaction_json( $property ) {
+		$input = array(
+			'properties' => array(
+				$property => array( 'http://target' ),
+			) );
+		return $input;
+	}
+
+	function test_create_reply() {
+		$input = $this->_test_create_interaction_json( 'in-reply-to' ); 
+		$post_content = Micropub_Render::generate_post_content( '', $input );
+		$this->assertEquals( '<p>In reply to <a class="u-in-reply-to" href="http://target">http://target</a>.</p>', $post_content );
+
+	}
+
+	function test_create_like() {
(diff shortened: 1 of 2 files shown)


tests/test_render.php
```diff
@@ -0,0 +1,182 @@
...
+		);
+		$post_content = Micropub_Render::generate_post_content( 'something', $input );
+
+		$this->assertEquals( "<p>Checked into <a class=\"h-card p-location\" href=\"http://place\">Place</a>.</p>\n" .
+			"<div class=\"e-content\">\nsomething\n</div>",
+			$post_content );
+	}
+
+/*
+	function check_create_content_html() {
```
snarfed: remove?

tests/test_render.php
```diff
@@ -0,0 +1,182 @@
...
+		$post = $this->check_create();
+		$this->assertEquals( 'HTML content test', $post->post_title );
+		// check that HTML in content isn't sanitized
+		$this->assertEquals( "<div class=\"e-content\">\n<h1>HTML content!</h1><p>coolio.</p>\n</div>", $post->post_content );
+	}
+ */
+	function _test_create_interaction_json( $property ) {
```
snarfed: minor nit: probably just `create_interaction_json`

tests/test_render.php
```diff
@@ -0,0 +1,182 @@
+<?php
+
+/** Unit tests for the Micropub Rendering class.
+ */
+
+class MicropubRenderTest extends WP_UnitTestCase {
+
+	public static function setUpBeforeClass() {
```
snarfed: i expect you can delete this

tests/test_render.php
```diff
@@ -0,0 +1,182 @@
...
+		WP_UnitTestCase::setUpBeforeClass();
+	}
+
+	public function setUp() {
```
snarfed: this too

snarfed (commented):
this absolutely makes sense. thank you for doing it.

i assume you also plan to delete the existing tests that these were copied from?

dshanske:
@snarfed Going to continue in this vein, but this one has been updated as requested if you want to approve.

tests/test_render.php
```diff
@@ -0,0 +1,176 @@
...
+		$content = '\n<h1>HTML content!</h1><p>coolio.</p>\n';
+		$input = array(
+			'properties' => array(
+				'content' => array( $content )
+			) );
+		$post_content = Micropub_Render::generate_post_content( 'something', $input );
+		$this->assertEquals( "<div class=\"e-content\">\n<h1>HTML content!</h1><p>coolio.</p>\n</div>", $post_content );
+	}
+
+	function create_interaction_json( $property ) {
```
snarfed: minor nit: remove `_json`, since this is after decoding the input data, ie not json specific.

tests/test_render.php
```diff
@@ -0,0 +1,176 @@
...
+		$this->assertEquals( '<p>Likes <a class="u-like-of" href="http://target">http://target</a>.</p>', $post_content );
+	}
+
+	function test_create_repost() {
+		$input = $this->create_interaction_json( 'repost-of' );
+		$post_content = Micropub_Render::generate_post_content( '', $input );
+		$this->assertEquals( '<p>Reposted <a class="u-repost-of" href="http://target">http://target</a>.</p>', $post_content );
+	}
+
+	function test_create_event_json() {
```
snarfed: ditto

snarfed (changes requested):
in general, this looks good. thanks!

one key point is that the new tests only cover the code paths _after_ deserializing the input form-encoded or json data. the old tests also tested those paths. so we should probably keep a few of the old ones around, notably one each of the `_post` (form-encoded) and `_json` tests, to keep testing those code paths.

dshanske:
With the paths, I was going to do one straight through in the next PR. Already in progress

snarfed:
great! you'll need at least two, for json vs form encoded.

generally i prefer to keep PRs atomic and not have any PR regress in functionality or test coverage. especially since in this case it's as easy as most not deleting a couple of the existing tests, especially since they do test these code paths fine as is, you don't really need to write new test(s). but if the next PR will include the replacement tests, sounds good.

snarfed (approved):

PR #143, opened 2018-07-25
Introduce REST API endpoint

dshanske:
This is still in process, but as the initial code functionality is there, and I'm still working on the test functionality(though you can see the start of it in the test_endpoint file). Wanted to ask for a review/feedback at this stage, as this is the most drastic of the changes I'll be doing out of all these PRs.

Will not be merging this commit till the unit tests are reimplemented using the new code. I also have some ideas for new tests that I'll likely put in once this is merged in. Specifically, user permissions test improvements, additional authorization tests with the new endpoint and related code, etc.

This PR switches the endpoint over to the REST functionality. To minimize the change here, it still uses the load_input to change over from the new endpoint into the mostly unchanged code.

The new unit tests will work similarly to the ones that are in the media endpoint, building the request using the WP_REST_Request class which simulates a payload, and the dispatch function which runs that payload through the endpoint, and testing the WP_REST_Response output.


pfefferle:
Tests are failing!

includes/class-micropub-endpoint.php
```diff
@@ -24,19 +24,15 @@ class Micropub_Endpoint {
...
-		// of ?micropub=endpoint, but that had problems. details in
-		// https://github.com/snarfed/wordpress-micropub/commit/d3bdc433ee019d3968be6c195b0384cba5ffe36b#commitcomment-9690066 )
-		add_filter( 'query_vars', array( $cls, 'query_var' ) );
-		add_action( 'parse_query', array( $cls, 'parse_query' ) );
-
 		// endpoint discovery
 		add_action( 'wp_head', array( $cls, 'micropub_html_header' ), 99 );
 		add_action( 'send_headers', array( $cls, 'micropub_http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'micropub_jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'micropub_jrd_links' ) );
```
snarfed: minor nit: indentation

includes/class-micropub-endpoint.php
```diff
@@ -117,14 +122,21 @@ protected static function check_scope( $scope ) {
...
 	 * Parse the micropub request and render the document
 	 *
-	 * @param int $user_id User ID for Authorized User.
+	 * @param WP_REST_Request $request.
 	 */
-	public static function post_handler( $user_id ) {
-		$status = 200;
+	public static function post_handler( $request ) {
+		$user_id = get_current_user_id();
+		if ( ! $user_id ) {
```
snarfed: looks like this lost the earlier `is_wp_error` check? do we need to add it back here?

```php
if ( is_wp_error( $user_id ) ) {
	static::handle_authorize_error( 401, $user_id->get_error_message() );
}
```

includes/class-micropub-endpoint.php
```diff
@@ -117,14 +122,21 @@ protected static function check_scope( $scope ) {
...
+			return new WP_Micropub_Error( 'unauthorized', 'Unauthorized', 401 );
+		}
+		$response = new WP_REST_Response();
+		static::load_input( $request );
+
 		$action = self::get( static::$input, 'action', 'create' );
 		if ( ! self::check_scope( $action ) ) {
-			static::error( 403, sprintf( 'scope insufficient to %1$s posts', $action ) );
+			return new WP_Micropub_Error( 'insufficient_scope', sprintf( 'scope onsufficient to %1$s posts', $action ), 403 );
```
snarfed: minor nit: typo, onsufficient

includes/class-micropub-endpoint.php
```diff
@@ -24,19 +24,15 @@ class Micropub_Endpoint {
...
-		add_filter( 'query_vars', array( $cls, 'query_var' ) );
-		add_action( 'parse_query', array( $cls, 'parse_query' ) );
-
 		// endpoint discovery
 		add_action( 'wp_head', array( $cls, 'micropub_html_header' ), 99 );
 		add_action( 'send_headers', array( $cls, 'micropub_http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'micropub_jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'micropub_jrd_links' ) );
 
+				// register endpoint
```
snarfed: minor nit: indentation

includes/class-micropub-endpoint.php
```diff
@@ -46,57 +42,66 @@ public static function get( $array, $key, $default = array() ) {
...
+				if ( 'action' === $key || 'url' === $key ) {
+					static::$input[ $key ] = $val;
+				} elseif ( 'h' === $key ) {
+					static::$input['type'] = array( 'h-' . $val );
+				} elseif ( 'access_token' === $key ) {
+					continue;
+				} else {
+					static::$input['properties']         = self::get( static::$input, 'properties' );
+					static::$input['properties'][ $key ] =
+					( is_array( $val ) && wp_is_numeric_array( $val ) )
```
snarfed: minor: for readability, either inline these two lines onto line 90 or indent them?

includes/class-micropub-endpoint.php
```diff
@@ -133,40 +145,41 @@ public static function post_handler( $user_id ) {
...
 			$uids[] = self::get( $syn, 'uid' );
 		}
+
 		$properties     = self::get( static::$input, 'properties' );
 		$synd_requested = self::get( $properties, 'mp-syndicate-to' );
 		$unknown        = array_diff( $synd_requested, $uids );
 
 		if ( $unknown ) {
-			static::error( 400, 'Unknown mp-syndicate-to targets: ' . implode( ', ', $unknown ) );
+			return new WP_Micropub_Error( 'insufficient_scope', sprintf( 'Unknown mp-syndicate-to targets: %1$s', implode( ', ', $unknown ) ), 400 );
```
snarfed: invalid_request, not insufficient_scope, right?

includes/class-micropub-endpoint.php
```diff
@@ -182,18 +195,19 @@ public static function post_handler( $user_id ) {
 				}
 			}
 			if ( ! $found ) {
-				static::error( 400, 'deleted post ' . $url . ' not found' );
+				return new WP_Micropub_Error( 'forbidden', sprintf( 'deleted post %1$s not found', $url ), 400 );
```
snarfed: invalid_request, not forbidden?

includes/class-micropub-endpoint.php
snarfed: also, probably worth unifying with the same code in `query_handler`.

includes/class-micropub-endpoint.php
```diff
@@ -249,12 +269,12 @@ private static function query_handler( $user_id ) {
...
 					break;
 				default:
-					static::error( 400, 'unknown query ' . static::$input['q'] );
+					return new WP_Micropub_Error( 'invalid_request', 'unknown query', 400, static::$input );
 			}
 		}
 
 		do_action( 'after_micropub', static::$input, null );
-		static::respond( 200, $resp );
+		return rest_ensure_response( $resp );
```
snarfed: interesting. is this useful? if so, should we do it at the end of `post_handler` too?

includes/class-micropub-endpoint.php
snarfed: oh, and we lost the code in `handle_authorize_error` that allows auth errors on localhost. that's useful. mind adding it back?

tests/test_endpoint.php
```diff
@@ -0,0 +1,227 @@
...
+		'h'         => 'entry',
+		'content'   => 'my<br>content',
+		'mp-slug'   => 'my_slug',
+		'name'      => 'my name',
+		'summary'   => 'my summary',
+		'category'  => array( 'tag1', 'tag4' ),
+		'published' => '2016-01-01T04:01:23-08:00',
+		'location'  => 'geo:42.361,-71.092;u=25000',
+	);
+		// JSON mf2 input
```
snarfed: minor nit: indentation. also lines 34, 54, etc.

tests/test_endpoint.php
```diff
@@ -0,0 +1,227 @@
+<?php
+/* Endpoint Tests */
+
+class Micropub_Endpoint_Test extends WP_UnitTestCase {
+
+
+	protected static $author_id;
+	protected static $subscriber_id;
+
+	// POST args
```
snarfed: can all this test data be reused from `test_micropub.php` instead of duplicated?

snarfed (commented):
this is great. i know i haven't felt like REST API was a priority in the past (and it's still not, per se), but i'm still very glad that we're switching to it. thank you for doing all this work!

tests/test_endpoint.php
dshanske: I intended to remove the test_micropub file in favor of the endpoint file as part of the split. So it will be gone in a future commit.

dshanske:
It isn't a priority, but I find it makes a lot of things easier.

includes/class-micropub-endpoint.php
snarfed: ping

tests/test_endpoint.php
snarfed: ping

tests/test_endpoint.php
snarfed: great! sounds good. mind adding a TODO?

includes/class-micropub-endpoint.php
snarfed: ping on all three of these

includes/class-micropub-authorize.php
```diff
@@ -25,6 +25,9 @@ class Micropub_Authorize {
 	// Array of Scopes
 	protected static $scopes = array();
 
+	// Error
+	protected static $error = null;
```
snarfed: this needs to get set to NULL somewhere at the beginning of request handling, right? otherwise, since it's static, when it gets set to an error in one request, it will keep that value across other future requests.

includes/class-micropub-authorize.php
```diff
@@ -35,8 +38,49 @@ public static function init() {
 		add_action( 'send_headers', array( $cls, 'http_header' ) );
 		add_filter( 'host_meta', array( $cls, 'jrd_links' ) );
 		add_filter( 'webfinger_user_data', array( $cls, 'jrd_links' ) );
+		add_filter( 'rest_index', array( $cls, 'rest_index' ) );
```
snarfed: heh, wow. i read the wordpress docs for `rest_index`, and i still have no idea what it's for. hopefully @pfefferle can help review some of the modern wordpress usage in this PR!

includes/class-micropub-authorize.php
```diff
@@ -35,8 +38,49 @@ public static function init() {
...
+
 		// The WordPress IndieAuth plugin uses priority 30
 		add_filter( 'determine_current_user', array( $cls, 'determine_current_user' ), 31 );
+		add_filter( 'rest_authentication_errors', array( $cls, 'rest_authentication_errors' ), 10 );
+		add_filter( 'rest_post_dispatch', array( $cls, 'return_oauth_error' ), 10, 3 );
+
+	}
+
+	public static function return_oauth_error( $result, $server, $request ) {
```
snarfed: this should probably be named something else, right?

includes/class-micropub-authorize.php
snarfed: ...more importantly, using long-lived state like this to affect control flow across multiple different parts of the code is hard to follow and error prone. we currently use eg `static::error()` to immediately stop and return an error at different points right now. i assume that no longer works because it's not the REST API way? can we find anything similar for REST? i'd much rather do that instead.

includes/class-micropub-endpoint.php
```diff
@@ -77,24 +113,14 @@ protected static function load_input( $request ) {
...
-				} elseif ( 'access_token' === $key ) {
-					continue;
-				} else {
-					static::$input['properties']         = self::get( static::$input, 'properties' );
-					static::$input['properties'][ $key ] =
-					( is_array( $val ) && wp_is_numeric_array( $val ) )
-					? $val : array( $val );
-				}
-			}
+				static::$input = self::form_to_json( $request->get_body_params() );
```
snarfed: `form_to_json` is static. does this need to be `static::form_to_json(`?

snarfed: same question with all of the existing `self::get(` calls, since it's static too...?

includes/class-micropub-authorize.php
dshanske: As noted, and will document, this is only for authorization. Otherwise, the default return message doesn't allow us to explain why it was rejecting, it will just say unauthorized. All Micropub errors are returned directly to the user.

includes/class-micropub-authorize.php
dshanske: Well, rest_index is just an index to allow individuals to know what services are available. Since we are providing indieauth authorization, we are just listing it in the index so people know it is an acceptable method.

includes/class-micropub-authorize.php
dshanske: micropub_error?

includes/class-micropub-endpoint.php
dshanske: Discussed elsewhere. Yes, PHP is weird.

dshanske:
@snarfed With your approval, I think this is ready to merge. I have additional tests planned. I have also done manual tests with two clients and Micropub.rocks

tests/test_endpoint.php
snarfed: ping

tests/test_endpoint.php
snarfed: ping

snarfed (approved):
looks great. thank you for your patience with all my questions and requests and ignorance! just a TODO and a couple indentation fixes left. feel free to merge once those are done!

PR #145, opened 2018-08-03
Dependency and Documentation Improvement

dshanske:
This starts the process of updating the README, as well as updates the dependencies, notes PHPUnit as a composer dependency so it could be installed as part of the development package, and moves error_log to an error function inside the endpoint, opening the door for future customization. It also removes a call to an activation hook that isn't there.


snarfed:
woo thank you! just fyi i'm away on a brief vacation, so it may take me a day or two to get to this.

dshanske:
I will continue to work on PRs.

snarfed (approved):
woo! thank you for docs! i don't have much (any) experience with travis or phpcs, so i'm pretty useless reviewing this. all i have is a typo. :P hey @pfefferle, any chance you could take a look?

includes/class-micropub-endpoint.php
```diff
@@ -49,8 +49,18 @@ public static function return_micropub_error( $response, $handler, $request ) {
 		return $response;
 	}
 
-
-
+	public static function log_error( $message, $name = '' ) {
```
pfefferle: If you set  `'Micropub'` as default for `$name`, you do not have to check in line 59.

includes/class-micropub-endpoint.php
```diff
@@ -49,8 +49,18 @@ public static function return_micropub_error( $response, $handler, $request ) {
...
+		if ( empty( $message ) ) {
+			return false;
+		}
+		if ( is_array( $message ) || is_object( $message ) ) {
+			$message = wp_json_encode( $message );
+		}
+		if ( ! is_string( $name ) || empty( $name ) ) {
+			$name = 'Micropub: ';
```
pfefferle: Replace `$name = 'Micropub: ';` with `$name = 'Micropub';` becaue the `: ` is set by the error logger in line 62

pfefferle:
Travis and PHPCS looks good to me... @dshanske Have you thought about an auto-releaser, like we use for the other IndieWeb plugins?

dshanske:
@pfefferle I thought about an autoreleaser. I also use a push script for several plugins as well. I figured I'd defer to snarfed.

dshanske:
If he's interested, I'd do it.

PR #148, opened 2018-08-09
Move permissions to be checked before all other actions

dshanske:
This PR just moves permissions to be checked before all other actions and enhances the scope check to check the scope against the user who is authorized to use it.

I'm thinking of adding something to the IndieAuth endpoint to not grant scopes that don't match user permissions in the first place, but IndieAuth.com doesn't do that.

It also moves setting scope and auth response into the permissions check functionality.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -87,7 +87,31 @@ class Micropub_Endpoint {
 		);
 	}
 
+
+	public static function get_action( $request ) {
+		$action = $request->get_param( 'action' );
+		return $action ? $action : 'create';
+	}
+
+	public static function load_auth() {
+		static::$micropub_auth_response = apply_filters( 'indieauth_response', static::$micropub_auth_response );
+		static::$scopes                 = apply_filters( 'indieauth_scopes', static::$scopes );
+	}
+
+	public static function get_client_id() {
+		return self::get( static::$micropub_auth_response, 'client_id', false );
+	}
+
+	public static function get_me() {
+		return self::get( static::$micropub_auth_response, 'me', false );
+	}
+
 	public static function check_query_permissions( $request ) {
+		self::load_auth();
+		$query = $request->get_param( 'q' );
+		if ( ! $query ) {
+			return new WP_Error( 'invalid_request', 'Missing Query Parameter', array( 'status' => 400 ) );
+		}
 		if ( ! current_user_can( 'read' ) ) {
 			return new WP_Error( 'forbidden', 'Unauthorized', array( 'status' => 403 ) );
 		}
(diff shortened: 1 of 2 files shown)


snarfed:
interesting! what's the motivation for moving the checks? it makes the code more complicated and duplicates the if/then tree for different actions, and complexity is the enemy of security (as is duplication :P), so especially for security related code, i'm not sure we'd want to do that unless there's a good reason.

dshanske:
That is where permissions are supposed to be checked. I am not sure it adds complexity per se. I see your point, but it puts all permission checks in the same place.

dshanske:
I had the idea of reorganizing the calls by action in future as well, maybe I should table this one for now.

snarfed:
> That is where permissions are supposed to be checked.

ah, true! the REST API `permission_callback`. got it. sounds good to me then.

> I had the idea of reorganizing the calls by action in future as well, maybe I should table this one for now.

ok! let me know if you still want me to review this.

dshanske:
If you are fine with this now, I can do the action changes in a different PR in future.

includes/class-micropub-endpoint.php
```diff
@@ -140,14 +166,39 @@ protected static function load_input( $request ) {
 	 * Check scope
 	 *
 	 * @param string $scope
+	 * @param id $user_id. If supplied will check user permissions
```
snarfed: it took me a while to understand the logic here, ie:
* if we can map the token to a user, check the scope against the user's permissions
* otherwise, assume we can do anything, and just check the token's scopes themselves

mind adding that to the comment here? or in your own words if i got it wrong?

tests/test_endpoint.php
```diff
@@ -199,6 +202,15 @@ public function test_create_post_without_create_scope() {
 		static::$scopes = array( 'post' );
 	}
 
+	public function test_create_post_subscriber_id() {
```
snarfed: thanks for the test!

tests/test_endpoint.php
```diff
@@ -199,6 +202,15 @@ public function test_create_post_without_create_scope() {
...
+		static::$scopes = array( 'create' );
+		$response       = $this->dispatch( self::create_form_request( static::$post ), static::$subscriber_id );
+		self::check( $response, 403, sprintf( 'user id %1$s cannot create posts', static::$subscriber_id ) );
+		// Set Back to Default
```
snarfed: minor: a better way to do this is in `setUp()`: https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/#shared-setup-between-related-tests . more reliable. (same w/lines 201-202, and maybe others.)

snarfed (approved):
thanks for the explanations! i get how this fits wordpress's REST structure better. looks great. some feedback but nothing blocking.

includes/class-micropub-endpoint.php
```diff
@@ -87,20 +87,48 @@ public static function register_route() {
...
+		$action = $request->get_param( 'action' );
+		return $action ? $action : 'create';
+	}
+
+	public static function load_auth() {
+		static::$micropub_auth_response = apply_filters( 'indieauth_response', static::$micropub_auth_response );
+		static::$scopes                 = apply_filters( 'indieauth_scopes', static::$scopes );
+	}
+
+	public static function get_client_id() {
```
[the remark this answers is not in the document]
dshanske: I want to use these to retrieve the client_id and me parameter as originally intended but I will save them for a PR where I actually do something with them.

dshanske:
@snarfed I made some modifications. I added tests for additional scope scenarios as well as adjusting the post handling calls of create, update, delete, and undelete to be a bit cleaner. Noticed that the return for insufficient scope was 403 instead of 401. 

Also, when adjusting the post handler, noticed an optimization for the get_posts call for trashed posts to only return the ids instead of the full post object.

snarfed:
looks good!

PR #149, opened 2018-08-11
Fix php notices

dshanske:
This should fix the PHP notices by checking for additional variables.

### DIFF
--- a/includes/class-micropub-render.php
+++ b/includes/class-micropub-render.php
@@ -9,20 +9,12 @@ add_filter( 'micropub_post_content', array( 'Micropub_Render', 'generate_post_co
  */
 class Micropub_Render {
 
-	public static function get( $array, $key, $default = array() ) {
-		if ( is_array( $array ) ) {
-			  return isset( $array[ $key ] ) ? $array[ $key ] : $default;
-		}
-		return $default;
-	}
-
-
 	/**
 	 * Generates and returns a post_content string suitable for wp_insert_post()
 	 * and friends.
 	 */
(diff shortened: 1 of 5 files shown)


includes/class-micropub-render.php
```diff
@@ -8,12 +8,21 @@
  * Micropub Render Class
  */
 class Micropub_Render {
+
+	public static function get( $array, $key, $default = array() ) {
```
snarfed: this is already duplicated in two other places. we should probably make it a single, normal top-level function to avoid that.

snarfed (changes requested):
otherwise looks good!

includes/class-micropub-render.php
dshanske: It will have to be prefixed then.

snarfed: ugh, right. so silly.

what do other plugins do about this? what does WordPress do? is their code just littered with `isset` checks? or do they structure it better somehow so they don't have to? or...?

dshanske: I don't think there is a better solution, but will look at structure

dshanske: I don't know how I ended up with conflicts, but here is my attempt at global functions.

snarfed:
thanks! so what was the conclusion on how other plugins and core handle possibly unset array accesses like this?

dshanske:
@snarfed Mostly isset with ternery.

dshanske:
@snarfed Do you want me to try another solution?

snarfed (approved):

Issue #150, opened 2018-08-14
Release Version 2.0 of Micrpub plugin


dshanske:
Version 2.0 is merely the name I'm using for all the changes thus far. Since I have decided I'm not going to pursue any further enhancements before releasing a new version, using this issue to track anything left to do before releasing this version as stable.

Asked @chrisaldrich , @jgmac1106, and @dougbeal if they would mind testing the current stable version for issues, as my tests with various Micropub clients seem to be working. Tests conducted with and without the IndieAuth plugin activated. 

Subject to that and any other blockers, want to share this with the world.

jackjamieson2:
A couple reports from my testing.

**1. When reposting from monocle or together, the following php notices occur:** 

[14-Aug-2018 16:46:50 UTC] Micropub Input: {"type":["h-entry"],"properties":{"repost-of":["https:\/\/jackjamieson.net\/511-2\/"]}}
 
14-Aug-2018 16:12:27 UTC] PHP Notice: Undefined index: content in/...{path}.../wp-content/plugins/wordpress-micropub-master/includes/class-micropub-endpoint.php on line 616 

[14-Aug-2018 16:12:27 UTC] PHP Notice: Undefined index: summary in/...{path}.../wp-content/plugins/wordpress-micropub-master/includes/class-micropub-endpoint.php on line 622

[14-Aug-2018 16:12:27 UTC] PHP Notice: Undefined index: content in/...{path}.../wp-content/plugins/wordpress-micropub-master/includes/class-micropub-render.php on line 79 

A repost is created, but it only consists of a link to the post. The content of the repost is otherwise empty.  I don't know if I've ever micropubbed a repost before so I'm not sure if this is the expected behaviour. 

**2. Posting from Quill with attached image**
Everything works, including the image, but my log includes an 404 message about the media endpoint

[14-Aug-2018 16:55:14 UTC] REST request: /micropub/1.0/media: {"q":"last"} 

[14-Aug-2018 16:55:14 UTC] REST result: : {"code":"rest_no_route","message":"No route was found matching the URL and request method","data":{"status":404}}

dshanske:
Re 2 above, https://indieweb.org/micropub_media_endpoint#Query_the_last_thing_uploaded - Basically, only Quill supports querying an endpoint. But right now, it generates an error because the endpoint doesn't support querying. I will change it and allow it to be queried, but it will return a proper error at this point, as not supporting last in this round.

miklb:
[...]

PR #152, opened 2018-08-17
Notices and Query

dshanske:
This fixes the errors noted in #150 removing the remaining PHP notices as well as fixing the error of the media endpoint having no query option. 

The micropub_post_content filter was completely ignoring that $post_content is passed through and replacing it with the content from properties, which defeats the purpose of a filter, so it now passes the content through to be altered.

However, this revealed a problem that replace occurred before delete, and therefore, it was overriding replace and setting things to null, so I swapped the two. In theory, if you are deleting and replacing something in the same transaction, which test_update does, you should delete first anyway.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -425,7 +425,7 @@ class Micropub_Endpoint {
 				);
 			}
 		}
-
+		// Delete was moved to before replace in versions greater than 1.4.3 due to the fact that all items should be removed before replacement
 		// delete
 		$delete = mp_get( $input, 'delete', false );
 		if ( $delete ) {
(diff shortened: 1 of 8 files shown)


snarfed:
thanks for the work here! these are all great fixes.

if the replace fix depends on replace now _happening_ after delete, please add a comment there saying so, and why. also, it sounds like we need at least a couple new tests, for the micropub_post_content filter and replace vs delete, if not others too.

finally, it looks like there are as many as four or more unrelated changes in this PR? PHP notices, media endpoint query, media endpoint permissions, post_content filter and replace vs delete. we've talked before about how closely related changes in the same PR is ok, but unrelated changes should generally be in separate PRs. please do that in the future.

thanks again!

dshanske:
@snarfed These are all related to the PHP Notices reported in #150, that's why I bracketed them in one single PR. But you are right. All reported together but technically separate issues. Will document and add the appropriate tests, as there was no test that will confirm the filter is working as expected...namely as a filter.

dshanske:
@snarfed I made some changes in the render testing, but also moved summary so it is part of the render class, which makes sense, as taking summary and setting it as the content is a rendering function, and shouldn't be in the endpoint itself.

snarfed (approved):
looks great. thank you!

PR #153, opened 2018-08-18
Add nag notice for insecure sites

dshanske:
If the site is not SSL this will add a notice telling people. For people who aren't concerned there is a constant they can add to disable the notice.

Ideally, no site should be sending tokens insecurely.

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -68,7 +68,7 @@ function micropub_not_ssl_notice() {
 	}
 	    ?>
     <div class="notice notice-warning">
-        <p>For security reasons you should use Micropub only on an HTTPS domain. You can disable this message be adding define( 'MICROPUB_DISABLE_NAG', '1' ); to your wp-config.php</p>
+        <p>For security reasons you should use Micropub only on an HTTPS domain.</p>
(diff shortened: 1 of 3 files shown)


micropub.php
```diff
@@ -58,3 +62,15 @@
 // Render Functions
 require_once plugin_dir_path( __FILE__ ) . 'includes/class-micropub-render.php';
 
+function micropub_not_ssl_notice() {
+	if ( is_ssl() || MICROPUB_DISABLE_NAG ) {
+		return;
+	}
+	    ?>
+    <div class="notice notice-warning">
+        <p>For security reasons you should use Micropub only on an HTTPS domain. You can disable this message be adding define( 'MICROPUB_DISABLE_NAG', '1' ); to your wp-config.php</p>
```
snarfed: minor nit, probably remove the "You can disable this..." sentence and replace it with something like "Your site is currently serving over HTTP." People who are able to define a custom PHP constant will know to look in the docs for that constant. :P

snarfed (approved):
...otherwise looks good!

PR #155, opened 2018-08-18
Split query into its own function

dshanske:
I want to be able to call this separately, so I split it into its own function.


includes/class-micropub-endpoint.php
```diff
@@ -349,6 +339,23 @@ public static function query_handler( $request ) {
 		return new WP_REST_Response( $resp, 200 );
 	}
 
+	public static function query( $post_id ) {
```
snarfed: mind adding a comment describing what this does, args and return value/type, etc?

snarfed (approved):
otherwise lgtm!

PR #156, opened 2018-08-18
If there is a token then do not pass control back to WordPress

dshanske:
This came up in regard to Omnibear, which is a browser extension Micropub client. Because the browser was logged into the WordPress site, cookie auth was being utilized and the individual was logged in, however, without scope as they didn't come from a token.

This change removes that check. Now, if there is no token, it will try cookie auth, but if there is a bad token, it will fail even if the person is logged in in that browser.

### DIFF
--- a/includes/class-micropub-authorize.php
+++ b/includes/class-micropub-authorize.php
@@ -150,7 +150,8 @@ class Micropub_Authorize {
 	}
 
 	/**
-	 * Validate the access token at the token endpoint.
+	 * Attaches to the determine_current_user filter and passes back the $user_id if there is no token.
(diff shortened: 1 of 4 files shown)


snarfed (commented):
huh. i think i understand this, but i'm not sure. any chance you can draft a unit test for it?

dshanske:
Actually, thinking about it, I can call the function without the entire chain

dshanske:
@snarfed This is what I came up with.

snarfed (approved):
lgtm with an explanation in the method docstring! discussed in irc.

PR #159, opened 2018-08-20
Reject cookie authentication

dshanske:
This attempts to reject all requests to the endpoint that might come over cookie authentication, to address more of the Omnibear issue(as the browser based extension is getting confused when someone is logged into the site in the same browser). I tried to simplify this a bit by failing if the auth response is empty.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -96,9 +96,11 @@ class Micropub_Endpoint {
 		}
 
 		// If there is no auth response this is cookie authentication which should be rejected
+		// https://www.w3.org/TR/micropub/#authentication-and-authorization - Requests must be authenticated by token
 		if ( empty( static::$micropub_auth_response ) ) {
 			return new WP_Error( 'unauthorized', 'Cookie Authentication is not permitted', array( 'status' => 401 ) );
 		}
+		return true;
 	}
 
 	public static function check_query_permissions( $request ) {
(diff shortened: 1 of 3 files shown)


includes/class-micropub-endpoint.php
```diff
@@ -90,22 +90,35 @@ public static function register_route() {
 	public static function load_auth() {
 		static::$micropub_auth_response = apply_filters( 'indieauth_response', static::$micropub_auth_response );
 		static::$scopes                 = apply_filters( 'indieauth_scopes', static::$scopes );
+		// Every user should have this capability which reflects the ability to access your user profile and the admin dashboard
+		if ( ! current_user_can( 'read' ) ) {
+			return new WP_Error( 'forbidden', 'Unauthorized', array( 'status' => 403 ) );
+		}
+
+		// If there is no auth response this is cookie authentication which should be rejected
```
snarfed: minor nit: mind explaining why?

snarfed (approved):
looks great! mind adding a test for this case? then feel free to merge!

includes/class-micropub-endpoint.php
dshanske: You shouldn't be able to log into Micropub without a token

dshanske: I probably should use the LOCAL AUTH variable

snarfed: > You shouldn't be able to log into Micropub without a token

right, sorry, i was hoping you could add that to the comment. maybe ideally with a link to the relevant part of the spec. thanks!

dshanske: Was planning on it. But wanted to make sure you got the meaning

PR #160, opened 2018-08-21
Add Commit to Travis per @pfefferle deployment script

dshanske:
Getting ready for simple deployment


snarfed (commented):
i don't know travis, so I'll defer to @pfefferle, but...

.travis.yml
```diff
@@ -74,6 +77,8 @@ before_install:
       bash bin/install-wp-tests.sh wordpress_test root '' localhost $WP_VERSION
       set +e
     fi
+after_success:
```
snarfed: this is sketchy. we should probably add it to the repo (and review it) instead.

.travis.yml
```diff
@@ -14,9 +14,12 @@ env:
   - WP_VERSION=latest WP_MULTISITE=0
   global:
   - WP_TRAVISCI=travis:phpunit
+  - SVN_REPO: https://plugins.svn.wordpress.org/micropub/
+  - secure: "lWqKSX3SPggddb0ass4pmep1GJ3g/l0iTpRTi+q8qaMlD9jFQ7bwMstXQ0QWPtRHqeX2pjbNTTol4Uw4j9djUdQm4XSOM+I7ki2bAOC0yPTy8sVeJ8uBy1cOI2Az2Ni+bYe4vg6ThhF9f9Q1rtb2Xs+suBZYCbUXP7tin+Z6sE3XMtWxmqaIIHOW4FW3z7ha0Dmxqj4fcqgWirEk8rNY55dt1phPfvWUKLbYdANemXnLz9dVf4aEQu/KfKGwlQFfFq4AILrtpeo3BkBIt6ZnU+AZeecyQSB0LEauphrF5fpKfPhsOksLUO/t6SVlWgSJ6K+36axj/OgqD2Ruo+DwYbHQCcVZ+BRJ2ESmch2hM1YZhzBwHkT21cKR8VCZkJkXtBoou2UjBfsw8SMy5M/taJjlJI9SjBl0ZRv/1sviXlEJovHCqg7yy85hX7kGVrULi1m+dDim96x/+YV3O7/VWZRi3mdwQ2lRBqkdpxqLAPnsgvFiNhrDyEKj1gT3G9sVlEpIrg3rw1KZDcP1eOO/DvchSIl3pW2uupN5fmCA+REFXYbLCTioKsfgOcRjPwUx24d4gPsYWhy7bfCOu1Kz9tZSA11Zbz0bVT5fB+awKcKaQkJg2s1JazaSrjBH/HSGBEXXIOXvGS+mUX/I+t2z1NzroHkmqQHWozXRl2Sh6Cs="
```
snarfed: is this a token for something? if so it shouldn't be in git anywhere, right? please revoke it, generate a new one, and put that token somewhere else. 😁

dshanske: https://docs.travis-ci.com/user/encryption-keys/ - Only Travis can decrypt it.

.travis.yml
aaronpk: I added it to my repo for Aperture because I didn't like the idea of running code from someone else's repo that might change at a moments notice. This is how the terrible npm hacks happen so I would like to avoid that.

.travis.yml
snarfed: oh wow, ok! please add a comment and docs link. thanks!

dshanske:
Going to defer until @pfefferle comments as this was originally his implementation.

.travis.yml
pfefferle: what user have you used for the commit?

dshanske: That would be my credentials right now. But snarfed wants to move to the Indieweb repo, so we should change to them

.travis.yml
dshanske: I think he's talking about the deployment script, but okay.

snarfed: (kidding! this one would be understandable, and not as painful as eg semantic linkbacks. i still vote for bundling though. 😁)

dshanske: After I merge the script, how do you feel about a release? Do we need more user feedback?

snarfed: i think we still have outstanding user reports of bugs and things not working, right? we should solicit more testers first, and then make sure all of those reports are fixed.

dshanske: I am waiting for information from some people

dshanske: But I want to be ready to go

pfefferle (dismissed):
I tried to sum up all the feedback:

* Change the commit user to indieweb:
  * change the user in the travis file (should I make a pull request to your branch or have you access to the indieweb 1password group?)
  * add the indieweb user as commiter for the WordPress.org directory
* Move the deploy script to the local bin folder
* Add `push.sh` to the `.svnignore` file

dshanske:
I've added the indieweb user to the readme, which once there is a push, it can push the entire repo. I'd suggest we retire the push.sh file.  I'll refactor this. I don't have access to the indieweb 1password group.

pfefferle:
@dshanske @snarfed don't forget to add the indieweb user as admin for the WordPress.org repo.

miklb:
does this still need that info updated in the deploy script?

dshanske:
I don't see any reason why we can't merge this now that we've addressed the local deployment issue. Any final thoughts?

dshanske:
As no objections noted, merged.

PR #162, opened 2018-08-22
Change config to filter

dshanske:
Oddly enough, there were issues @miklb had issues with the plugin because it was trying to load both the built-in auth and the IndieAuth. So I decided that a more dynamic way of doing this might make more sense, and retiring the configuration variable.

Since the Micropub spec now says that auth must be in the form of OAuth2 or the IndieAuth variant, it shouldn't support anything called MICROPUB_LOCAL_AUTH. So, the new version disables the auth in favor of it being handled externally. I also added to the documentation.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -32,7 +32,7 @@ Supports the following [scope](https://indieweb.org/scope) parameters requested 
 
 
 ### Filters and hooks 
-Adds seven filters:
+Adds six filters:
 
 `before_micropub( $input )`
 
@@ -48,11 +48,6 @@ Called to generate the list of `syndicate-to` targets to return in response to a
 `micropub_query( $resp, $input )`
 
 $resp defaults to null. If the return value is non-null, it should be an associative array that is encoded as JSON and will be returned in place of the normal micropub response.
-
-`enable_micropub_auth( $boolean )`
-
-If this filter returns true the authentication functions built into the plugin are enabled. By default, this is disabled if the IndieAuth Plugin is installed. By default it will return
-true which loads the built-in IndieAuth client.
 
 `indieauth_scopes( $scopes )`
 
@@ -109,9 +104,13 @@ WordPress has a [whitelist of file extensions that it allows in uploads](https:/
 
 For reasons of security it is recommended that you only use this plugin on sites that implement HTTPS.
 
-Supports the full OAuth2/IndieAuth authentication and authorization flow. Defaults to IndieAuth.com. Custom auth and token endpoints can be used by overriding the `MICROPUB_AUTHENTICATION_ENDPOINT` and `MICROPUB_TOKEN_ENDPOINT` endpoints or by setting the options `indieauth_authorization_endpoint` and `indieauth_token_endpoint`.
+Supports the full OAuth2/IndieAuth authentication and authorization flow. Defaults to IndieAuth.com. Custom auth and token endpoints can be used by overriding the `MICROPUB_AUTHENTICATION_ENDPOINT`
+and `MICROPUB_TOKEN_ENDPOINT` endpoints or by setting the options `indieauth_authorization_endpoint` and `indieauth_token_endpoint`.
 
 If you want to use your own site as an IndieAuth endpoint, you can activate the IndieAuth plugin which is recommended but not required. You can disable the authentication in favor of an alternative plugin.
+This can be done by removing the loading of the auth flow or setting MICROPUB_LOCAL_AUTH to 1.
+
+`remove_action( 'plugins_loaded', 'load_micropub_auth', 20 );` 
 
 If the token's `me` value matches a WordPress user's or author post URL, that user will be used. If there is only one site author that will be matched otherwise.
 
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -94,15 +106,13 @@ WordPress has a [whitelist of file extensions that it allows in uploads](https:/
...
+For reasons of security it is recommended that you only use this plugin on sites that implement HTTPS.
 
-If the token's `me` value matches a WordPress user's or author post URL, that user will be used. If there is only one site author that will be matched otherwise.
+Supports the full OAuth2/IndieAuth authentication and authorization flow. Defaults to IndieAuth.com. Custom auth and token endpoints can be used by overriding the `MICROPUB_AUTHENTICATION_ENDPOINT` and `MICROPUB_TOKEN_ENDPOINT` endpoints or by setting the options `indieauth_authorization_endpoint` and `indieauth_token_endpoint`.
 
-Alternatively, you can set `MICROPUB_LOCAL_AUTH` to 1 to disable the plugin's authorization function, for example if you want authorization to be done by WordPress or another plugin. It will also 
-be disabled if the IndieAuth plugin is installed.
-
-Finally, for ease of development, if the WordPress site is running on `localhost`, it logs a warning if the access token is missing or invalid and still allows the request.
```
snarfed: does this PR remove this feature? or was it removed before? it's useful, and i think harmless; i'd like to keep it.

readme.md
```diff
@@ -206,6 +215,7 @@ into markdown and saved to readme.md.
 * Ensure compliance with Micropub spec
 * Update composer dependencies and include PHPUnit as a development dependency
 * Add nag notice for http domains and the option to diable with a setting
+* `MICROPUB_LOCAL_AUTH` is now deprecated in favor of a filter
```
snarfed: this PR does more than deprecate it; it removes it entirely. that's fine, it's not a compatibility concern, just checking on language.

readme.md
```diff
@@ -44,12 +44,23 @@ Called during the handling of a Micropub request. The content generation functio
...
 `micropub_syndicate-to( $synd_urls, $user_id )`
 
-Called to generate the list of `syndicate-to` targets to return in response to a query. Returns `$synd_urls`, an array, possibly modified.
-
+Called to generate the list of `syndicate-to` targets to return in response to a query. Returns `$synd_urls`, an array, possibly modified. This filter is empty by default
 `micropub_query( $resp, $input )`
 
 $resp defaults to null. If the return value is non-null, it should be an associative array that is encoded as JSON and will be returned in place of the normal micropub response.
 
+`disable_micropub_auth( $boolean )`
```
snarfed: mind mentioning explicitly what the arg and return value types are for these new filters?

snarfed (changes requested):
huh. ok. @miklb were you defining `MICROPUB_LOCAL_AUTH`? i'd like to be sure this is the cause of your problem before we do this. mind testing with this PR?

> Since the Micropub spec now says that auth must be in the form of OAuth2 or the IndieAuth variant, it shouldn't support anything called MICROPUB_LOCAL_AUTH. So, the new version disables the auth in favor of it being handled externally. I also added to the documentation.

technically, the spec doesn't say anything about a `disable_local_auth` filter any more than `MICROPUB_LOCAL_AUTH`. :P

i don't mind switching to a filter, but as is, this PR breaks anyone currently using `MICROPUB_LOCAL_AUTH`. maybe reconsider? or at least keep `MICROPUB_LOCAL_AUTH` compatibility and remove it in the next release? (you call it "deprecated" in the release notes, but this PR removes it entirely. deprecating would keep it. :P)

miklb:
I will test this PR. I do not have any reference to local in my WP config.

dshanske:
No one to my knowledge is using the setting. But somehow it was tripping for miklb. I thought as we don't need it, removal waa the best option.

readme.md
dshanske: It replaces using it with a filter, which means more in the plugin territory than user territory

snarfed: ...but also removes the existing built in behavior, and requires users to write PHP code (ie add a filter handler) to get it back. at minimum we'd need to document that...but i'd actually just like to keep the existing behavior, if that's ok.

mind adding it back in a separate PR? (since we'd probably only merge this one if it fixes @miklb's problem, which seems uncertain.)

feel free to grab me on IRC to talk more easily!

dshanske:
@snarfed I added some items here to try and address the variable issue. Wanted your thoughts. As for @miklb, he did prompt me to look at that line, but for some reason, he was tripping it despite not having it set.

dshanske:
@snarfed Seem to have made a mistake. I'm not sure what I missed here but the tests are failing, which suggests that I was right and something I had set about that definition isn't working. Will have to fix later.

miklb:
this latest change results in expected behavior authenticating with Quill including ability to post a note and syndicate.

💯

snarfed:
hey, that's great news! thank you for testing @miklb!

snarfed:
@dshanske sounds like the delayed check in your latest commit is what fixed the problem. that's great!

given that, the filter is probably unnecessary and overkill for a simple setting. mind switching it back to the old constant?

dshanske:
I thought of a way to have the best of both worlds

dshanske:
@snarfed I realized that by loading the auth on a hook, I accomplished the same goal, so the documentation now reflects that.

snarfed:
thank you so much @dshanske, this looks great!

Issue #172, opened 2018-11-04
Plugin protects access to all Wordpress wp-json API endpoints


lostfocus:
When the plugin is active all wp-json endpoints are returning a 401 error if called without a token, even the public ones  which isn't quite the expected and/or wanted behaviour.

(Example: `/wp-json/wp/v2/posts` which should be accessible without a token.)

lostfocus:
I'm pretty sure that's somewhere in `determine_current_user` https://github.com/indieweb/wordpress-micropub/blob/2.0.0/includes/class-micropub-authorize.php#L158 - calls to a public API endpoint usually do not have an auth header or a token.

dshanske:
Oddly, I'm seeing this now after the authorize piece moved in 2.0.1 to fix a different problem.When it shouldn't be loading.

lostfocus:
That could be something with the order of hooks and filters.

dshanske:
I'm working on it.

snarfed:
ugh, this bit me too. thanks @lostfocus!

@dshanske i'm definitely looking forward to the fix. the current status is pretty painful, since i can only have webmentions or micropub work, not both, so this seems pretty urgent. do you have a recommended workaround? or just expediting your fix?

dshanske:
Resolved in Version 2.0.2

PR #174, opened 2018-11-05
Don't throw an error when there is no auth header and no token.

lostfocus:
Not ideal, but this way it just returns the `$user_id` which is probably 0 at that point. This should solve #172 - at least it did for me.


lostfocus:
(I also failed at github. I'm pretty sure the merge commit isn't supposed to be in this pull request. Sorry.)

dshanske:
I'm maintaining two versions of this code. The other version doesn't have this problem. And the difference is that error message. I knew I fixed this problem.

https://github.com/indieweb/wordpress-indieauth/blob/master/includes/class-indieauth-authorize.php#L83

dshanske:
Don't worry about it. It will be fixed

PR #177, opened 2018-11-13
suppress PHP `strpos(): Empty needle` warning

snarfed:
when running outside of a web server, e.g. in wp-cli.

```
PHP Warning: strpos(): Empty needle in /usr/home/ryancb/public_html/w/wp-content/plugins/micropub/includes/class-micropub-authorize.php on line 166
```


dshanske:
Sorry, never even occurred to me as a use case.

PR #183, opened 2019-01-04
fix formatting with missing `

chrisaldrich:


pfefferle:
But the markdown is generated from the readme.txt, can you also file a merge request for the readme.txt?

PR #189, opened 2019-03-08
Parse geo URIs

dshanske:
Goal of this PR is to parse all geo URIs into mf2 objects.

The latest version of Indigenous embeds h-cards into geo URIs, so decided this was a better move.

@gRegorLove Thanks for volunteering to have a look at this.


gRegorLove (approved):
lgtm

PR #198, opened 2019-04-09
Fix issue with media not getting attached to post

### LINKED ISSUE #173: OwnYourGram media didn't get attached to post
Upload works fine, but the connection between media and post didn't work. I'l monitor this further to see if it was just a fluke or if it is a bug.

jjeaton:
wp_update_post requires the 'ID' parameter. http://codex.wordpress.org/Function_Reference/wp_update_post

I was able to get this to work with a local WP install and local version of OwnYourGram.

Fixes #173

### DIFF
--- a/includes/class-micropub-media.php
+++ b/includes/class-micropub-media.php
@@ -346,7 +346,7 @@ class Micropub_Media {
 			// Attach media to post
 			wp_update_post(
 				array(
-					'post_ID'     => $id,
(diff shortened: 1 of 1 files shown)


lostfocus:
👍🏻Thanks. I just put it on my server and it works for me.

dshanske:
I will see about merging this and releasing an update

PR #217, opened 2020-02-05
Add filters

dshanske:
This adds a filter to satisfy the request to custom post type and to customize taxonomies.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -437,6 +437,13 @@ class Micropub_Endpoint {
 			}
 		}
 		$args = static::mp_to_wp( static::$input );
+
+		// Allow Filtering of Post Type
+		$args['post_type'] = apply_filters( 'micropub_post_type', 'post', static::$input );
+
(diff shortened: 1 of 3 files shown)


includes/class-micropub-endpoint.php
```diff
@@ -437,6 +437,13 @@ private static function create( $user_id ) {
 			}
 		}
 		$args = static::mp_to_wp( static::$input );
+
+		// Allow Filtering of Post Type
+		$args['post_type'] = apply_filters( 'micropub_post_type', 'post', static::$input );
```
pfefferle: is this really correct? shouldn't it be something like

```$args['post_type'] = apply_filters( 'micropub_post_type', 'post', static::$input['post_type'] );```

???

pfefferle: ah sorry, misread it...

`post` is the param and `static::$input` is only the context.

dshanske: Yes, the request was for a way to set a custom post type or a custom taxonomy based on input parameters. So two specific filters.

Issue #226, opened 2020-07-19
Make IndieAuth Plugin a Dependency to Use This Plugin


dshanske:
https://chat.indieweb.org/dev/2020-07-14/1594761683886700

I would have to agree that we should be pushing local IndieAuth over remote IndieAuth by making hosting your local endpoint the default.

The solution to this seems to be to implement remote indieauth as a hidden feature in the IndieAuth plugin...which is currently a PR since January, and anyone who wishes it can activate it. 

For the Micropub plugin, will be working on a PR that will add an upgrade notice to install the Indieauth plugin, and after a sufficient time, remove the code entirely.

@snarfed As you were one of the people who prefers usage of indieauth.com, wanted your input on this.

aaronpk:
To clarify, the idea is to simplify this plugin by making it rely on the IndieAuth plugin for access token generation and verification? Essentially this plugin would assume the IndieAuth plugin exists and use it to verify access tokens in the requests?

if the IndieAuth plugin has the ability to then rely on remote endpoints like indieauth.com (although I still don't understand why you'd want to do that), this plugin wouldn't even see the difference right? Assuming this plugin asks the IndieAuth plugin to verify access tokens, it's up to the IndieAuth plugin to decide how to verify them whether that's it checking the local database or asking a remote token endpoint.

That seems like the best of both worlds and would simplify this plugin a lot.

I would also like to emphasize that my personal interest in this is getting more people to install the IndieAuth plugin instead of relying on indieauth.com. It's easier to tell someone to install two plugins than it is to tell them how to set up rel=me links on their home page in order to use IndieAuth.com. Also I am planning on eventually shutting that service down.

aaronpk:
To clarify, the idea is to simplify this plugin by making it rely on the IndieAuth plugin for access token generation and verification? Essentially this plugin would assume the IndieAuth plugin exists and use it to verify access tokens in the requests?

if the IndieAuth plugin has the ability to then rely on remote endpoints like indieauth.com (although I still don't understand why you'd want to do that), this plugin wouldn't even see the difference right? Assuming this plugin asks the IndieAuth plugin to verify access tokens, it's up to the IndieAuth plugin to decide how to verify them whether that's it checking the local database or asking a remote token endpoint.

That seems like the best of both worlds and would simplify this plugin a lot.

[...]

PR #227, opened 2020-07-19
IndieAuth Dependency

dshanske:
This would address #226, and would make IndieAuth(the plugin) or some future alternative IndieAuth plugin a dependency of this plugin.


dshanske:
Realized I didn't check the tests...I have to remove the checking of scopes in favor of checking of user capabilities. The latest version of IndieAuth the plugin will be removing user capabilities if not in the token.

PR #230, opened 2020-08-13
fix readme markdown...

pfefferle:
...update WordPress version and added update-assets workflow

### DIFF
--- a/.github/workflows/update-assets.yml
+++ b/.github/workflows/update-assets.yml
@@ -1,0 +1,17 @@
+name: Plugin asset/readme update
+on:
+  push:
+    branches:
+    - master
+jobs:
+  master:
+    name: Push to master
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: WordPress.org plugin asset/readme update
+      uses: 10up/action-wordpress-plugin-asset-update@master
+      env:
+        SVN_PASSWORD: ${{ secrets.SVN_PASSWORD }}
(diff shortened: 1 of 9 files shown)


pfefferle:
@dshanske the first test seems broken. I tried to fix it, but I can't run the tests locally. Can you help me here?

pfefferle:
Might be this change https://github.com/indieweb/wordpress-micropub/blob/trunk/micropub.php#L32 ? Is it a good idea to only load the classes if IndieAuth is installed? If so, we should require the classes in the tests.

dshanske:
I have a dummy version of the file included in bootstrap...your update took it out.

pfefferle:
Ah, I see! Re-added!

pfefferle:
I had to change the bootstrap file, to run under osX.

dshanske:
I didn't realize it didn't...it was the standard one generated at the time by wp-cli

dshanske:
Wonder if I need to check every bootstrap

pfefferle:
@dshanske can you help me fix the test, to merge the changes?

dshanske:
I got the testing working, but there is still an error on one...I'm not sure why it never came up before. Something new that is causing the terms not to delete?

dshanske (approved):
The changes are good, but somehow it caused some tests to fail. I assume because of the environment changes. I can't figure out why the test to delete categories isn't working. I've tried multiple variations of it.

pfefferle:
The text to markdown script is somehow broken... what about using the grunt scripts?

PR #238, opened 2020-09-09
Do Not Store Built-In Properties

dshanske:
This PR eliminates the storage of all mp- passed as these are commands, not properties.  It also stops storing published, updated, summary, and name as these are now solely stored in the WordPress post. To accommodate the fact that there is no timezone offset in WordPress, this is stored.


dshanske:
There is an issue here I need to fix. The published/updated dates are failing because I'm generating them based on the current time, so it is off by a second, which it isn't otherwise. Have to figure out how to adjust.

PR #244, opened 2021-03-13
This Moves Unit and Lint Testing from Travis to GitHub Actions

dshanske:
It also excludes build files from appearing in releases.


pfefferle:
Awesome! ❤️

dshanske:
@pfefferle I have more tweaks for this in future. But it's a start.

PR #257, opened 2021-09-28
Published/Updated are singular (non-array) fields

Ruxton:
Looking for [0] is breaking me posting old content. Published is supposed to be a singular (non-array) field when present.

Happy to re-write this to array check first if this is gonna cause problems somewhere.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -701,7 +701,7 @@ class Micropub_Endpoint extends Micropub_Base {
 		}
 
 		if ( isset( $props['published'] ) ) {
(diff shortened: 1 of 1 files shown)


dshanske:
@Ruxton https://micropub.spec.indieweb.org/#json-syntax It is an array. There is no singular property in JSON syntax. 

The only place there is an example in the spec that shows it is https://micropub.spec.indieweb.org/#example-21

Now, in form encoded it does allow a single non-array value, but if you look, the plugin converts form-encoded to json to standardize everything.

Ruxton:
My bad, was dealing with some old stuff that was sending published as singular and when I read the spec I didn't see it as such.

PR #258, opened 2021-09-29
Uploaded URLs filenames wont include query strings

Ruxton:
Ensures that when sent a URL with query strings (like instagram) the filename saved doesnt include the query string.


dshanske:
WordPress would use wp_parse_url, not parse_url. 

https://www.php.net/manual/en/features.file-upload.post-method.php

You also don't need to access the array, parse_url and wp_parse_url have a second parameter. 

It should be wp_parse_url( $url, PHP_URL_PATH )

PR #261, opened 2021-12-05
Add more comprehensive tests for post status determination

jeremyfelt:
This introduces test coverage for what I believe to be the possible combinations for the `visibility` and `post-status` properties when `Micropub_Endpoint::post_status()` is used to set post status.

The tests flagged an issue where if `visibility` is set to `public` and there is no `post-status` property, then `null` would be returned an an error generated. It seems safe to assume a default post status when `visibility` is set to `public`, so I added another commit to help tests pass.

I originally looked into this issue and thought (🤦🏻) the issue was a string/array mix up. After digging further, I see now that this plugin requires an array value for both `visibility` and `post-status` even though the spec seems to imply a string. Other portions of the plugin check for a numeric array in addition to the string. This seems like something that could be done here, but I wanted to check before going any further.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -646,7 +646,10 @@ class Micropub_Endpoint extends Micropub_Base {
 			}
 			return 'draft';
 		}
-		// Execution will never reach here
+
+		// If visibility is public and no post-status is specified,
+		// return the default post status value.
+		return self::default_post_status();
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


dshanske:
I'm going to review this further. I might turn to a bit more normalization.

dshanske (approved):

dshanske:
I'm verifying thoughts, but the plugin takes the form-encoded input and converts it to JSON to match the JSON encoded input, which should have no single properties.

dshanske:
I've committed a function that ensures even if a property in JSON syntax comes in without an array, it is converted so everything is standardized.

PR #285, opened 2023-03-02
Stop including visible text in reply contexts

snarfed:
...since they go inside since they go inside e-content, which webmention recipients use as the reply text.

https://chat.indieweb.org/wordpress/2023-02-27#t1677522099115700

### DIFF
--- a/includes/class-micropub-render.php
+++ b/includes/class-micropub-render.php
@@ -47,13 +47,24 @@ class Micropub_Render {
 					$val['name'] = $val['url'];
 				}
 				if ( isset( $val['url'] ) ) {
-					$lines[] = sprintf(
-						'<p>%1s <a class="u-%2s" href="%3s">%4s</a>.</p>',
-						$verbs[ $prop ],
-						$prop,
-						$val['url'],
-						$val['name']
-					);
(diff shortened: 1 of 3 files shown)


snarfed:
The PHP 5.6 failure looks unrelated. Does the code maybe no longer support PHP 5.6?

```
PHP Parse error:  syntax error, unexpected '[' in /home/runner/work/wordpress-micropub/wordpress-micropub/includes/class-micropub-endpoint.php on line 719
```

snarfed:
Confirmed, the last merged PR (#282) did the same thing, passed on PHP 7.2 but failed on 5.6: https://github.com/indieweb/wordpress-micropub/actions/runs/3175919981

So we'll probably want to either raise the required PHP version or fix that incompatibility, but either way that's unrelated to this PR, so hopefully we can still merge this as is!

dshanske:
I need to review the test process to see what introduced it

snarfed:
Sure! I've moved that to #286, to clarify that it's separate and shouldn't block merging this PR.

dshanske (approved):
I don't have any objection to this, but I don't use this version of the code, so I'm not sure if I should get another opinion

Issue #294, opened 2023-11-09
Plugin always adds e-content div to post content. Should it?


snarfed:
Hi @dshanske et al! I'm seeing an odd behavior with HTML content: two `<div class="e-content">`s, one inside the other. I assume that's not expected? For example, this JSON request:

```json
{
  "type": ["h-entry"],
  "properties": {
    "name": ["foo"],
    "category": ["reply"],
    "content": [{"html": "well <em>ok</em> is <a href=\"http://yahoo.com\">that so</a> <a href=\"http://bar.org\">http://bar.org</a>"}],
  }
}
```

results in this HTML:

```html
<div class="entry-content clear">
<div class='e-content'><a class="u-in-reply-to" href="https://micropub.spec.indieweb.org/"></a></p>
<div class="e-content">
well <em>ok</em> is <a href="http://yahoo.com">that so</a> <a href="http://bar.org">http://bar.org</a>
</div>
</div>
<div class="syndication-links"></div>
</div><!-- .entry-content -->
```

I'm guessing because of this code?

https://github.com/indieweb/wordpress-micropub/blob/9bccb0bdc501a1dca9441b713811248e82f43191/includes/class-micropub-render.php#L96-L100

`e-content` is probably up to the theme, right? Should the plugin be injecting this extra `e-content` div?

janboddez:
Just FYI: What I do in my theme (or rather, plugin, but a theme could do the same) is I check if there's a `.e-content` in the post content already and only add the class if there isn't. (For some post types, it makes sense to have a, e.g., reply context outside of `.e-context`, and since---in my case---both are part of my post content I can't just add an overall `e-content` class. Hence the check.)

But I also completely override the Micropub post content in a filter callback, so ... That could work too, at least temporarily. Just remove the `div` in a filter.

dshanske:
There's a setting that disables it entirely which my Post Kinds plugin turns on... I have some ideas I might pursue though.

snarfed:
FWIW, I debugged this a bit more just now, and it's not specific to HTML content or JSON input, evidently the plugin always does this when it creates posts. Eg a form-encoded micropub request with `content=fooey` results in this content HTML:

```html
<div class="e-content">
fooey
</div>
```

snarfed:
(I don't actually know what should happen here! Just adding info for now.)

snarfed:
[...]

Issue #211, opened 2019-12-29
Switch to dynamic rendering


dshanske:
Regardless of the considerations in how mf2 plugins should relate, this plugin should stop encoding microformats into the post content and add these on the fly. 

@snarfed, @pfefferle I need your input on this. 

My proposal is to mark all existing posts made by Micropub on upgrade by adding a meta key. That will have them processed the current way.

For new posts, post_content will only contain e-content, and the other properties will be added on the fly. 

This will be disabled if Post Kinds handles rendering. 

You already couldn't effectively use the editor for editing Micropub posts without manually editing the HTML. With Post Kinds you could, and with Gutenberg you still cannot.

So, this sets up for more flexibility on what we might do in future. I just want help ensuring nothing gets messed up.

snarfed:
🎉 🚢

dshanske:
> 🎉 🚢

A few minor things before the version release to do. Then wait for the fireworks.

PR #297, opened 2023-12-25
Dynamic Rendering

dshanske:
This is an intermediate step to add dynamic rendering. It keeps pre-2.4.0 posts using the static rendering, but everything new, that has the version tag within the Micropub auth response would use dynamic rendering unless either the Post Kinds plugin is enabled or some other unknown plugin changes the micropub_dynamic_render to true.

The only other plugin that amends the post content is @janboddez 's IndieBlocks, which converts the incoming Micropub content to blocks.

If this solution gets positive feedback as a solution my next step will be to enhance the rendering code to address all open issues. The idea being that this will render a simplified version of the data, Post Kinds will take that over if it is enabled, as will Indieblocks or a theme by removing the dynamic render filter or adding to the new filter.

### DIFF
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -988,6 +988,7 @@ class Micropub_Endpoint extends Micropub_Base {
 		if ( $micropub_auth_response || ( is_assoc_array( $micropub_auth_response ) ) ) {
 			$args['meta_input']                           = mp_get( $args, 'meta_input' );
 			$args['meta_input']['micropub_auth_response'] = wp_array_slice_assoc( $micropub_auth_response, array( 'client_id', 'client_name', 'client_icon', 'uuid' ) );
+			$args['meta_input']['micropub_auth_response']['version'] = micropub_get_plugin_version();
 		}
 		return $args;
 	}
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -1073,41 +1073,6 @@ class Micropub_Endpoint extends Micropub_Base {
 		return $args;
 	}
 
-	/**
-	 * Returns the mf2 properties for a post.
-	 */
-	public static function get_mf2( $post_id ) {
-		$mf2  = array();
-		$post = get_post( $post_id );
-
-		foreach ( get_post_meta( $post_id ) as $field => $val ) {
-			$val = maybe_unserialize( $val[0] );
-			if ( 'mf2_type' === $field ) {
-				$mf2['type'] = $val;
-			} elseif ( 'mf2_' === substr( $field, 0, 4 ) ) {
-				$mf2['properties'][ substr( $field, 4 ) ] = $val;
-			}
-		}
-
-		// Time Information
-		$published                      = micropub_get_post_datetime( $post );
-		$updated                        = micropub_get_post_datetime( $post, 'modified' );
-		$mf2['properties']['published'] = array( $published->format( DATE_W3C ) );
-		if ( $published->getTimestamp() !== $updated->getTimestamp() ) {
-			$mf2['properties']['updated'] = array( $updated->format( DATE_W3C ) );
-		}
-
-		if ( ! empty( $post->post_title ) ) {
-			$mf2['properties']['name'] = array( $post->post_title );
-		}
-
-		if ( ! empty( $post->post_excerpt ) ) {
-			$mf2['properties']['summary'] = array( $post->post_excerpt );
-		}
-
-		return $mf2;
-	}
-
 	/* Takes form encoded input and converts to json encoded input */
 	public static function form_to_json( $data ) {
 		$input = array();
--- a/includes/class-micropub-endpoint.php
+++ b/includes/class-micropub-endpoint.php
@@ -1003,8 +1003,7 @@ class Micropub_Endpoint extends Micropub_Base {
 	 */
 	public static function store_mf2( $args ) {
 		// Properties that map to WordPress properties.
-		// TODO: We need to still store content because the plugin adds markup to the content stored.
-		$excludes = array( 'name', 'published', 'updated', 'summary', 'updated' );
+		$excludes = array( 'name', 'published', 'updated', 'summary', 'updated', 'content', 'visibility' );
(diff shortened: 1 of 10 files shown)


dshanske:
@janboddez Would appreciate your input as well. 

This PR is only a proof of concept so far, and I tried to come up with a backward compatible solution.

dshanske:
Related #211 #294

snarfed:
Whoa! Awesome! This is a _huge_ deal. Thank you so much for tackling it!

I'll refrain from reviewing the code itself, my head has been out of this codebase and WordPress in general for a long time, but I have tons of questions! I'll start with just a few:
* How do you plan to render content? Eg if a post has an `in-reply-to` or a `bookmark-of`, what will the resulting markup look like? Will those parts be outside `e-content`?
* This doesn't affect how post-meta is rendered, specifically whether it has mf2, right?

dshanske:
> Whoa! Awesome! This is a _huge_ deal. Thank you so much for tackling it!
> 
> I'll refrain from reviewing the code itself, my head has been out of this codebase and WordPress in general for a long time, but I have tons of questions! I'll start with just a few:
> * How do you plan to render content? Eg if a post has an `in-reply-to` or a `bookmark-of`, what will the resulting markup look like? Will those parts be outside `e-content`?
> * This doesn't affect how post-meta is rendered, specifically whether it has mf2, right?

For now it calls the same code it was using. If I merge this, the next PR would improve and rewrite that code.

includes/class-micropub-endpoint.php
```diff
@@ -988,6 +988,7 @@ public static function store_micropub_auth_response( $args ) {
 		if ( $micropub_auth_response || ( is_assoc_array( $micropub_auth_response ) ) ) {
 			$args['meta_input']                           = mp_get( $args, 'meta_input' );
 			$args['meta_input']['micropub_auth_response'] = wp_array_slice_assoc( $micropub_auth_response, array( 'client_id', 'client_name', 'client_icon', 'uuid' ) );
+			$args['meta_input']['micropub_auth_response']['version'] = micropub_get_plugin_version();
```
pfefferle: Isn't it better to have that in a top level meta field, to maybe use it for migrations? Otherwise a query on serialized data is really hard!

dshanske:
> > Whoa! Awesome! This is a _huge_ deal. Thank you so much for tackling it!
> > I'll refrain from reviewing the code itself, my head has been out of this codebase and WordPress in general for a long time, but I have tons of questions! I'll start with just a few:
> > 
> > * How do you plan to render content? Eg if a post has an `in-reply-to` or a `bookmark-of`, what will the resulting markup look like? Will those parts be outside `e-content`?
> > * This doesn't affect how post-meta is rendered, specifically whether it has mf2, right?
> 
> For now it calls the same code it was using. If I merge this, the next PR would improve and rewrite that code.

That's definitely an endorsement of this approach. I experimented a bit with the idea of including an mf2 parser to try an interpret what is marked up inside. Also, the next version will remove storing the content property in meta, so it can use the presence of the content property to undo the changes.

janboddez:
Oh, IndieBlocks currently uses the `micropub_post_content` filter to, under certain conditions, parse in Like, Reply, etc. blocks (and basically remove whatever markup was generated by this plugin anyway).

So, as long as that hook is called the same way, I'm good. (If it ever gets deprecated, I'm going to have to fall back to some lower-level WordPress hook, which may prove tricky.)

May have to prevent Micropub from re-rendering the content I saved, though. Should be no more than a `remove_filter()` call? I could add that in any time (maybe even before this gets released).

dshanske:
> Oh, IndieBlocks currently uses the `micropub_post_content` filter to, under certain conditions, parse in Like, Reply, etc. blocks (and basically remove whatever markup was generated by this plugin anyway).
> 
> So, as long as that hook is called the same way, I'm good. (If it ever gets deprecated, I'm going to have to fall back to some lower-level WordPress hook.)
> 
> May have to prevent Micropub from re-rendering the content I saved, though. Should be no more than a `remove_filter()` call? I could add that in any time (maybe even before this gets released).

The micropub_post_content filter will stay as it is. In this case, all you would have to do is add a simple line.

add_filter( 'micropub_dynamic_render',  '__return_false`, 99, 2 ); Or since the 2nd argument will be $post...

Built in by default it will default to true if there is a version property in the micropub_auth_response, which I'll be adding from now on, or if there is an mf2_content property, which will be used prior to this version.

includes/class-micropub-endpoint.php
dshanske: Good point. I'll be revisiting this soon. We added the auth response for debugging, maybe time to reconsider it for new posts.

pfefferle (approved):
lgtm (but not tested)

readme.md
```diff
@@ -224,11 +227,12 @@ into markdown and saved to readme.md.
 
 ## Changelog ##
 
-### 2.4.0 (202x-xx-xx) ###
+### 2.4.0 (2024-xx-xx) ###
```
asuh, line 230: Should this date be filled in before deployment?

includes/class-micropub-endpoint.php
```diff
@@ -1003,8 +1004,7 @@ public static function store_micropub_auth_response( $args ) {
 	 */
 	public static function store_mf2( $args ) {
 		// Properties that map to WordPress properties.
-		// TODO: We need to still store content because the plugin adds markup to the content stored.
-		$excludes = array( 'name', 'published', 'updated', 'summary', 'updated' );
+		$excludes = array( 'name', 'published', 'updated', 'summary', 'updated', 'content', 'visibility' );
```
asuh: `'updated'` listed twice here?

readme.md
dshanske: Before release, yes, but not before merge. I always set the date before I release the version.

snarfed:
Thanks again for doing this, it's a huge step forward for microformats2 support in WordPress. I skimmed the code, it seems fine, but I don't live in WordPress, so honestly that doesn't mean much. Still, the facts that it's only dynamic going forward, IndieBlocks-compatible, and preserves the existing rendering for now are all great. If @pfefferle's on board, then I am too. Exciting!

PR #299, opened 2024-03-13
Add IndieAuth as Plugin Dependency

pfefferle:
See: https://make.wordpress.org/core/2024/03/05/introducing-plugin-dependencies-in-wordpress-6-5/

### DIFF
--- a/micropub.php
+++ b/micropub.php
@@ -7,6 +7,7 @@
  * Author: IndieWeb WordPress Outreach Club
  * Requires at least: 4.9.9
  * Requires PHP: 5.6
+ * Requires Plugins: indieauth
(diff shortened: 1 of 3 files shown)


snarfed (approved):
LGTM!

dshanske:
I was going to ask about backcompat, but I think we have it covered.

pfefferle:
What do you mean with backcompat? It will be ignored by all versions below 6.5, so it should be fine.

dshanske:
> What do you mean with backcompat? It will be ignored by all versions below 6.5, so it should be fine.

Yes, and I'm still checking for the presence of the plugin class, but I'm wondering about the long-term.

PR #300, opened 2024-03-13
Add a WordPress.org Plugin Repo Icon

pfefferle:
See: https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/#plugin-icons


snarfed (commented):
Looking at https://developer.wordpress.org/plugins/wordpress-org/plugin-assets/ , should this be in `assets/`, not `.wordpress-org/`?

pfefferle:
@snarfed the plugin uses the 10up GitHub Workflows and these default to `.wordpress-org`: 

* https://github.com/10up/action-wordpress-plugin-asset-update?tab=readme-ov-file#optional-environment-variables
* https://github.com/10up/action-wordpress-plugin-deploy?tab=readme-ov-file#optional-environment-variables

snarfed (approved):

PR #310, opened 2026-01-02
Add WordPress hooks extractor workflow

pfefferle:
## Summary

- Add `extract-wp-hooks.yml` workflow using [akirk/extract-wp-hooks](https://github.com/akirk/extract-wp-hooks) action
- Automatically extracts filters and hooks documentation to the GitHub wiki
- Remove filters and hooks section from readme (now auto-generated)

The workflow runs on push to trunk when PHP files change, or can be triggered manually.


dshanske:
This one Iove... I should add this to all the projects I work on

PR #313, opened 2026-01-11
Avoid duplicate e-content wrapper when theme supports microformats2

### LINKED ISSUE #294: Plugin always adds e-content div to post content. Should it?

pfefferle:
## Summary
- Only wraps content in an `e-content` div when the theme doesn't already provide one
- Checks explicit theme support via `current_theme_supports( 'microformats2' )`
- Adds a `micropub_theme_supports_microformats2` filter for themes/plugins that output an `e-content` wrapper without declaring theme support
- The conditional wrapper lives entirely in the dynamic rendering path introduced in #297

An earlier revision tried to auto-detect `e-content` in the rendered page output at shutdown; that was dropped because detection only ever ran on Micropub posts — where the plugin itself had just injected the wrapper — so the first page view would cache a false positive and permanently disable the wrapper for every theme.

Note: this only affects dynamically rendered posts. Pre-2.4.0 posts have the wrapper baked into `post_content` and need a separate fix.

Fixes #294


dshanske:
How does this integrate with #297 where I tried to switch to dynamic rendering?

pfefferle:
@dshanske so this is already fixed but simply not released?

dshanske:
@pfefferle Not exactly, the new rendering code only worked for new posts, not for old ones..as in pre 2.4.0

pfefferle:
But it seems to be still an issue!?

/cc @snarfed

pfefferle:
This integrates naturally with #297's dynamic rendering. The `theme_supports_microformats2()` check is called inside `generate_post_content()`, which is only invoked when dynamic rendering is active (`should_dynamic_render()` returns true).

So when dynamic rendering is enabled:
- If the theme declares `microformats2` support OR the cached detection found `e-content` in the template → content is rendered without the extra wrapper
- Otherwise → content is wrapped in `<div class="e-content">`

The detection runs lazily on the first singular page view (when the option getter is called and no cache exists), then the result is cached. Cache is cleared on theme switch.

Also addressed the Copilot feedback:
- Store `theme_supports_microformats2()` result in a local variable to avoid calling twice
- Use `has_action()` to prevent adding shutdown hook multiple times  
- Set autoload to false for the option
- Check if `preg_replace` returns null before continuing
- Improved docblock documentation
- Renamed test function for clarity

pfefferle:
@dshanske I rebased this onto trunk and reworked it to answer your question about #297:

- The conditional wrapper now lives entirely inside the dynamic rendering path from #297 — `generate_post_content()` is only reached via the `the_content` filter. So this is a refinement of dynamic rendering, not a parallel mechanism: the renderer simply skips the `e-content` div when the theme already provides one.
- While reworking I dropped the output-buffer auto-detection from the first revision. It could never work reliably: it only ran on Micropub posts, where the plugin itself had just injected `e-content`, so the first page view would cache a false positive and disable the wrapper for every theme (and without an active output buffer at shutdown, `ob_get_contents()` returns nothing at all). It's now just `current_theme_supports( 'microformats2' )` plus a `micropub_theme_supports_microformats2` filter for plugins/themes that add their own wrapper without declaring support — the check @janboddez described in #294.
- You're right that this doesn't help pre-2.4.0 posts — those have the wrapper baked into `post_content`. I'd rather track that separately (strip-on-render or a migration) than mix it into this PR.

/cc @snarfed

PR #320, opened 2026-05-02
Return errors from wp_insert_post / wp_update_post failures

### LINKED ISSUE #319: Errors return HTTP 201 status
Hi @pfefferle @dshanske et al! I noticed an odd quirk recently: some error requests return HTTP 201. I'm guessing that should be 400 instead? Example:

```sh
$ curl -v -H "Authorization: Bearer $token" -d "h=entry&repost-of=$url&post-status=draft" https://snarfed.org/wp-json/micropub/1.0/endpoint
...
< HTTP/2 201 
...
{
  "ID": {
    "ID": 0,
    "data": {
      "error": "empty_content",
      "error_description": "Content, title, and excerpt are empty."
    },
    "filter": "raw",
    "headers": [],
    "status": 200
  },
  "meta_input": {
    "mf2_post-status": [
      "draft"
    ],
    "mf2_repost-of": [
      "https://bsky.app/profile/mackinac.social/post/3mktckulfys2k"
    ],
    "mf2_type": [
      "h-entry"
    ],
    "micropub_auth_response": {
      "client_icon": "https://secure.gravatar.com/avatar/947b5f3f323da0ef785b6f02d9c265d6?s=96&d=blank&r=g",
      "client_id": "https://ryandc36.wpcomstaging.com/wp-admin/",
      "client_name": "indie-respond browser extension"
    },
    "micropub_version": {
      "version": "2.5.0"
    }
  },
  "post_author": 1,
  "post_status": "draft",
  "post_type": "post",
  "post_url": false,
  "tax_input": null
}
```

pfefferle:
## Summary

Fixes #319.

When `wp_insert_post` or `wp_update_post` returned a `WP_Error`, `check_error()` wrapped it as a Micropub `Error` and stored it in `$args['ID']`, but `handle_create()` / `handle_update()` returned the `$args` array as if the operation had succeeded. The endpoint then responded with HTTP 201 and a body that included the error nested inside the `ID` field, plus `post_url: false` and `Location: false`.

- Propagate the Micropub error from `handle_create()` and `handle_update()` when insertion fails, so `create_item()` returns a proper error response.
- Change the default status in `micropub_wp_error()` from 200 to 400. A `WP_Error` always indicates failure, so a 200 default never made sense; this also surfaces the correct status for `wp_insert_post`'s built-in errors (like `empty_content`) which carry no `status` data.

## Test plan

- [x] New regression test `test_create_returns_error_when_wp_insert_post_fails` reproduces the exact scenario from #319 (uses the `wp_insert_post_empty_content` filter to force the error). Fails on `trunk` (returns 201), passes with the fix (returns 400 with `error: empty_content`).
- [x] Full PHPUnit suite passes (86 tests, 337 assertions).
- [x] PHPCS clean on changed files.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -299,7 +299,7 @@ if ( ! function_exists( 'micropub_wp_error' ) ) {
 	function micropub_wp_error( $error ) {
 		if ( is_wp_error( $error ) ) {
 			$data   = $error->get_error_data();
-			$status = isset( $data['status'] ) ? $data['status'] : 200;
+			$status = isset( $data['status'] ) ? $data['status'] : 400;
 			if ( is_array( $data ) ) {
 				unset( $data['status'] );
 			}
--- a/includes/rest/class-endpoint-controller.php
+++ b/includes/rest/class-endpoint-controller.php
@@ -522,6 +522,11 @@ class Endpoint_Controller extends \WP_REST_Controller {
 		}
 
 		$this->insert_post( $args );
+
+		if ( \is_micropub_error( $args['ID'] ) ) {
+			return $args['ID'];
+		}
+
 		$this->default_file_handler( $args['ID'] );
 
 		return $args;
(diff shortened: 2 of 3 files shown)


dshanske:
This makes sense to me

dshanske (approved):

